PR #103, opened 2014-10-09
Parallel installer for Neovim

junegunn:
This pull request implements parallel installer using Neovim's job control.

(/cc @justinmk)
#### Asynchronous

Due to the nature (or limitation?) of job control, this version of parallel installer is asynchronous. (Polling doesn't work since it blocks main thread.) You can freely move your cursor, edit other buffers while installation is ongoing. Which is cool and all but it's inconsistent with the current behavior, and it's not anymore possible to write something like the following:

``` sh
# The installer is asynchronous, so this will exit immediately
nvim +PlugUpdate +qa
```

Being asynchronous also means that the user can close the plug window or switch tabs anytime during the installation. If the window is closed, we have to make sure that the running jobs are correctly terminated. Currently I'm calling `jobstop()` function to abort tasks, but I've noticed that the git processes are not immediately killed.

So we definitely need more tests but unfortunately I don't think it's possible to write test cases for asynchronous behavior using Vader.vim.
#### ~~Unable to detect errors~~

As far as I know, the current API of job control does not expose the exit status of the process.

_EDIT_: Implemented with a workaround
#### Task timeout not implemented

I didn't bother to implement task timeout.
#### Windows support

I don't think `jobstart(name, 'sh', ['-c', args])` would run on Windows. :unamused: 
Does Neovim support Windows?

### DIFF
--- a/plug.vim
+++ b/plug.vim
@@ -643,7 +643,7 @@ function! s:is_managed(name)
 endfunction
 
 function! s:names(...)
-  return filter(keys(g:plugs), 'stridx(v:val, a:1) == 0 && s:is_managed(v:val)')
+  return sort(filter(keys(g:plugs), 'stridx(v:val, a:1) == 0 && s:is_managed(v:val)'))
 endfunction
 
 function! s:update_impl(pull, force, args) abort


justinmk:
> I don't think jobstart(name, 'sh', ['-c', args]) would run on Windows.

@junegunn Probably you already know, but normally `sh -c ...` should not be needed for neovim jobs. Instead just call `jobstart(name, [arg1, arg2, ...])`.

> The only missing feature that seemed relevant was error detection, but I managed to find a simple, albeit naive, workaround for it.

There's movement on this here: https://github.com/neovim/neovim/pull/1844

junegunn:
> normally sh -c ... should not be needed for neovim jobs

@justinmk I see but the argument to jobstart can be [quite lengthy](https://github.com/junegunn/vim-plug/blob/16642/plug.vim#L820-L822) (chained commands), so it was easier for me to just use `sh -c`.

Thanks for the update!

benekastah:
Instead of passing `sh` or `bash` when I need to run a program with args as a string, I've used `&shell`. That's one step in the direction of windows support, but I have no idea if windows shells can accept a program as an argument with `-c` like bash and related shells can. It's an idea, anyway.

starcraftman:
@benekastah In general, I assume by Windows support you mean cmd.exe the default shell. In that case, no such animal. To start on Windows they use the very logical `/` instead of `-` to signal flags. Secondly, I believe by default you can pass cmd.exe an arbitrarily long string but not -c.

You can see the flags at > http://www.microsoft.com/resources/documentation/windows/xp/all/proddocs/en-us/cmd.mspx?mfr=true

On a related note, I may help with Windows support soonish, but working on my python support fork right now.

justinmk:
Passing the arguments as a list to jobstart avoids shell escaping problems, and will avoid cmd.exe completely on windows.

PR #143, opened 2015-01-07
Fix #139 - Use git fetch+merge instead of pull

### LINKED ISSUE #139: PlugUpdate does not handle new branches
Ran into a bug when trying to pass the `branch` option with a newly-created branch. It seems that `PlugUpdate` does not fetch new branches because I get the following error:

```
error: pathspec 'issue23' did not match any file(s) known to git.
```

 I had to manually run `git fetch` in the plugin's repo folder before executing `PlugUpdate` for it to work.

junegunn:


junegunn:
This should fix #139. One thing I'm worried about is the use of `--ff-only` option, which was introduced on Git 1.6.6.

junegunn:
Further research revealed that vim-plug is using the features introduced in 1.6.5, `--recursive` option for `git clone` and `git submodule`. 1.6.5 was released on October 2009.

vheon:
@junegunn I would go for it, I mean, 5 years is a lot of time software speaking. In the remote chance that someone will complain, at least we'll have some sense of what we could consider the bare minimum.

junegunn:
@vheon Thanks! I was a little afraid that this change might break some corner cases that we're not aware of it. But, what can we do? At least, all the tests pass, let's go for it.

PR #144, opened 2015-01-12
Add <nowait> argument for buffer-local mappings

pgdouyon:
I ran into some conflicts with some of my global mappings and buffer-local mappings in Vim-Plug's preview window.  This should prevent forcing the user to wait for the 'timeout' setting to expire when there is a conflict so that Vim-Plug feels a lot snappier.

### DIFF
--- a/plug.vim
+++ b/plug.vim
@@ -564,14 +564,14 @@ function! s:prepare()
     silent %d _
   else
     call s:new_window()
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks, but `<nowait>` option was added on Vim 7.3 (`:help added-7.3`), but vim-plug should work on older versions of Vim as well. We could look at `v:version`, but I don't think it's really worth it.

PR #147, opened 2015-01-17
Simplify neovim check

Pyrohh:
The semantics are the same, but I think it's clearer what's happening.
This was made possible in https://github.com/neovim/neovim/commit/da2906f28e1f5ee4cbff5fa2e0c0ec88777b838a and https://github.com/neovim/neovim/commit/037ca796a0782fc03409dd7f9500113937ed32b7.


junegunn:
Thanks!

PR #156, opened 2015-01-29
Don't force filetype plugin and indent

tommcdo:
Use previously set plugin and indent flags when re-enabling filetype detection. Calling `filetype on` will leave `plugin` and `indent` flags unchanged, allowing a user to specify these value (anywhere before `plug#end()`).

### DIFF
--- a/plug.vim
+++ b/plug.vim
@@ -212,7 +212,7 @@ function! plug#end()
   endfor
 
   call s:reorg_rtp()
-  filetype plugin indent on
+  filetype on
   if has('vim_starting')
     syntax enable
   else


starcraftman:
@tommcdo The travis build is broken. Unlikely it will be accepted until fixed.

vheon:
@tommcdo whats a use case for this? I mean I always want ftdetect, ftplugin and indent file...

tommcdo:
@vheon I personally hate indent. And, from a more purist standpoint, a plugin manager should not have side effects like this.

@starcraftman when I get some more time I'll look into why the build is failing.

junegunn:
Thanks. But every change at this point should be backward-compatible. There are many users, including myself, who have organized their vimrcs with the assumption that vim-plug does this for you, and I think it's an acceptable default considering that most people do not disagree on it and Vundle and NeoBundle state in their document that the line is "required".

If you're going to add an optional argument to the function (e.g. `call plug#end({'indent': 0, 'syntax': 0})`) that's fine but I can't accept this as the default behavior.

junegunn:
And how about simply adding `filetype indent off` after `plug#end()` call?

tommcdo:
@junegunn Adding `filetype indent off` works fine for now. If I get a moment I may implement the suggestion from your previous comment.

junegunn:
@tommcdo Okay, since `filetype indent off` works just as expected without any significant overhead (the same can be said about `syntax off`), I don't think we need to add the optional parameter. I'm closing the PR. Thanks.

tommcdo:
Having an argument like `plug#end({'plugin': 1, 'indent': 0})` suggests that `indent` will be turned off, regardless of its prior value. Since these are off by default in Vim, how about something like `plug#end(['plugin'])` instead? It would be strange to turn it on early in your vimrc and then instruct plug to turn it off. Instructing plug to leave it unchanged might be sufficient (and perhaps more sensible).

The behaviour would be like this:

Invocation | Effect
--- | ---
`plug#end()` or `plug#end(['plugin', 'indent'])` | `filetype plugin indent on`
`plug#end(['indent'])` | `filetype indent on`
`plug#end(['plugin'])` | `filetype plugin on`
`plug#end([])` | `filetype on`

junegunn:
@tommcdo Thanks for the detailed suggestion. Your suggestion is very sensible and complete, so I've given some thought about it, but I concluded that it's not going to be really useful. I get that you don't like `indent`, it's understandable, but how many users would want to exclude `plugin`? It's not realistic. We can implement the scheme for the completeness' sake, but this new option is probably not going to be used by many and can introduce unnecessary confusion. So I'd say we should stick to the workarounds previously mentioned and keep it simple.

tommcdo:
Yeah, I think that's a reasonable conclusion. The workaround works perfectly fine.

Thanks!

PR #163, opened 2015-02-08
Remove `--progress` option from fetch and clone.

sangh:
This breaks on old versions of git (at least 1.7.0.4, but I didn't go
look to see where this option is introduced).  I think it can just be
removed because you can't see it in the `:PlugUpdate` screen.


junegunn:
Thanks for the suggestion. We can't simply remove `--progress` since the code is also used on neovim, which shows the progress on-the-fly. I'll fix it.

sangh:
Yup, seems fine, thanks!

PR #165, opened 2015-02-12
Python parallel plugin management.

starcraftman:
Log:
* Main difference from ruby is inside Command class, notably it
poll/sleeps on the subprocess to check output & timeout. Not overly taxing CPU.
* Otherwise, I mainly just mirrored logic into an OOP style.
* Note that due to GVim freeze, disabling use on Windows. Also, likely
other issues.
---------------------------------------

I've squashed and cleaned up the commit and everything seems to work.
One thing I should note, is that unlike ruby, the update_python function blocks (i.e. :PlugInstall when using Python).
I think it may be a difference in the Python embedded interpreter that I can't fix. We can make a ticket here to look at it, minor issue.

Hope it is to your liking @junegunn . I've mainly testing on Ubuntu 14.04, someone like @vheon maybe should do a complete re-run of everything on Mac. 

IMPORTANT: For testing purposes, if you have +ruby & +python it always chooses the ruby version. To test either manually false the s:ruby flag or compile a python only version.

### DIFF
--- a/plug.vim
+++ b/plug.vim
@@ -73,6 +73,7 @@ let s:plug_tab = get(s:, 'plug_tab', -1)
 let s:plug_buf = get(s:, 'plug_buf', -1)
 let s:mac_gui = has('gui_macvim') && has('gui_running')
 let s:is_win = has('win32') || has('win64')
+let s:py2 = has('python') && !s:is_win
 let s:ruby = has('ruby') && (v:version >= 703 || v:version == 702 && has('patch374'))
 let s:nvim = has('nvim') && !s:is_win
 let s:me = resolve(expand('<sfile>:p'))
@@ -745,7 +746,7 @@ function! s:update_impl(pull, force, args) abort
     \ 'pull':    a:pull,
     \ 'force':   a:force,
     \ 'new':     {},
-    \ 'threads': (s:ruby || s:nvim) ? min([len(todo), threads]) : 1,
+    \ 'threads': (s:py2 || s:ruby || s:nvim) ? min([len(todo), threads]) : 1,
     \ 'bar':     '',
     \ 'fin':     0
   \ }
@@ -754,13 +755,17 @@ function! s:update_impl(pull, force, args) abort
   call append(0, ['', ''])
   normal! 2G
 
-  if s:ruby && s:update.threads > 1
+  if (s:py2 || s:ruby) && !s:nvim && s:update.threads > 1
     try
       let imd = &imd
       if s:mac_gui
         set noimd
       endif
-      call s:update_ruby()
+      if s:ruby
+        call s:update_ruby()
+      else
+        call s:update_python()
+      endif
     catch
       let lines = getline(4, '$')
       let printed = {}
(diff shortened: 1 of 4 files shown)


plug.vim
```diff
@@ -1332,6 +1698,8 @@ function! s:upgrade()
       endif
     elseif s:ruby
       call s:upgrade_using_ruby(new)
+    elseif s:py2
+      call s:upgrade_using_python(new)
     else
       return s:err('curl executable or ruby support not found')
```
vheon: This should mention python as well I think.

starcraftman: @vheon Good catch, I did a line-by-line merge from my squash onto a new branch for fast forward. Missed this.

starcraftman: @vheon I think I read your comment too quickly, what do you mean 'mention python'? In README?

vheon: `'curl executable or ruby support not found'` this string should mention python.

vheon: I did a comment on a line but I don't know, since it wasn't a line you added, if github attached my comment on that line or at the file instead.

starcraftman: Ah, the error msg. I was only looking at the green. Classic tunnel vision. :)

junegunn:
Thanks! I'll try to review the code during the weekend.

starcraftman:
@junegunn No problem. If you have questions feel free to ping me. I think I'm done with it now, after improving the tracing/logging. 

GLog can be useful if you want to trace particular things during execution.

plug.vim
```diff
@@ -968,6 +973,364 @@ while 1 " Without TCO, Vim stack is bound to explode
...
+class CmdTimedOut(Exception):
+  pass
+class CmdFailed(Exception):
+  pass
+class InvalidURI(Exception):
+  pass
+class Action(object):
+  INSTALL, UPDATE, ERROR, DONE = ['+', '*', 'x', '-']
+
+class GLog(object):
```
junegunn: I'm okay with this now, but I would like to consider removing GLog in the future once we're confident about the stability of the code.

junegunn:
I just skimmed through the code. Looks solid, great job! I'm testing it right now, it works pretty well, but on here I'll focus on the differences between the ruby version and the python version.

---

When you `:PlugUpdate`, Ruby version shows `+` sign for the plugins that are being cloned, and `*` for the pre-existing ones. Python version always shows `*`.

---

The CPU consumption of Python version is significantly higher. On my Macbook Air, Vim process occupies 100% when Python version is running. Do you think you can improve it? Both Ruby installer and Neovim installer employ some strategies to limit the rate of buffer updates that turned out to be very effective in limiting CPU usage of Vim itself.

https://github.com/junegunn/vim-plug/blob/master/plug.vim#L1079-L1082
https://github.com/junegunn/vim-plug/blob/master/plug.vim#L826-L828

---

There are some differences in timeout handling. Ruby version [increases the time limit](https://github.com/junegunn/vim-plug/blob/master/plug.vim#L1066) on every retrial (constant backoff), but python version keeps using the same value. And Ruby version gives [visual indication of timeouts](https://github.com/junegunn/vim-plug/blob/master/plug.vim#L1097), but Python version doesn't. Try installing/updating several dozens of plugins with `let g:plug_timeout = 1`, you'll notice the difference.

plug.vim
starcraftman: @junegunn I don't mind either way. When GLog is disabled (default) I don't think it contributes to any slowdown as it returns almost immediately. I figured it would be handy later, if someone reports a py bug I can just have em pastebin the log.

starcraftman:
@junegunn Differences number 1 & 3 are fairly straightforward. Just oversights during my port. 

2 maybe be a bit more tricky. I mean I knew it was a CPU hog doing that polling and just relying on scheduler. I guess I'll do some testing and try to get it lower to the ruby/nvim level. My only reservation is I think the buffer will look slightly sluggish/jerky by comparison. I guess most people prefer their cpus not get pinged.

plug.vim
junegunn: @starcraftman Yes, it would be useful at this stage. But at some point, it will be stable enough that it's no longer necessary (there hasn't been any bug reports on Ruby installer for quite a while) and I'd prefer to have smaller code base.

junegunn:
@starcraftman Note that I ran into the same problem of high CPU overhead when I first implemented live progress update of Ruby installer. I noticed that the overhead slowed down the whole installation process almost by a factor of two on my old mid-2011 Macbook Air, which I'm still using. So I had to make a trade-off and limit the rate of the update.

starcraftman:
@junegunn I believe I've addressed the issues. 1 & 3 behave like the ruby version now. I also added interrupt ability that I had forgotten. So ctrl + c will interrupt a long package.

Re CPU, I've used the method from ruby that reduces log calls. Seems acceptable to me. Adequate CPU reduction?

junegunn:
@starcraftman Thanks for the quick follow up. I can confirm `+` issue is fixed, CPU reduction is adequate, and I'm seeing visual indication of timeouts. Great!

There's a few more things I've noticed while testing it with extremely short timeout (`let g:plug_timeout = 1`). I'm cloning 34 plugins and updating 20 plugins here.

---

Ruby versions [removes the directory](https://github.com/junegunn/vim-plug/blob/54fc8a4fc/plug.vim#L1165-L1167) when git-clone (not pull, only clone) is interrupted by timeout. But Python version doesn't. Not cleaning up the directory from the previous attempts makes the subsequent trial all fail.

- Ruby
ruby

- Python
python

---

And with `g:plug_timeout` set to 1, I *occasionally* run into the following case where the installer terminates without processing every plugin. Any idea?

screen shot 2015-02-15 at 9 53 35 am

---

Interrupting parallel installer is not as responsive as doing so in Ruby installer. I had to hit CTRL-C several times to make it completely stop.

junegunn:
The second case is not readily reproducible. But does happen from time to time.
(The same condition, timeout 1 sec, 34 clones and 20 pulls)
https://gist.github.com/junegunn/e0cc2ba5054d0535dd4f

junegunn:
See http://junegunn.kr/plug_logs.tgz as well

starcraftman:
@junegunn Very interesting problems. Re number 1, that is strange, must be some permission problem. 

I've got no clue on number 2, looks very strange. The line complaining about is just a delete on the curbuf. Will investigate. I'm getting 404 on the plug_logs.

Re number 3, I see why. In ruby, a single ctrl+c cancels all threads, I just made it cancel a single Plugin job. Will raise an issue to fix that.

junegunn:
>  I'm getting 404 on the plug_logs.
Oops, sorry. Restored the file.

> must be some permission problem

I don't think it's related to permissions. `attempt_cmd()` does not `rmtree` on retrials.

starcraftman:
@junegunn Thanks for files, and yes. I see now. I was doing cleanup only after all retries not in between. Holdover from before I did the timeouts.

renstrom:
Ran your Python code through [PEP8](https://github.com/jcrocholl/pep8) (ignore `E111` which is "indentation is not a multiple of four").

```bash
$ pep8 --ignore E111 vim-plug.py
vim-plug.py:20:17: E226 missing whitespace around arithmetic operator
vim-plug.py:23:1: E302 expected 2 blank lines, found 1
vim-plug.py:23:29: E701 multiple statements on one line (colon)
vim-plug.py:24:1: E302 expected 2 blank lines, found 0
vim-plug.py:24:27: E701 multiple statements on one line (colon)
vim-plug.py:25:1: E302 expected 2 blank lines, found 0
vim-plug.py:25:28: E701 multiple statements on one line (colon)
vim-plug.py:26:1: E302 expected 2 blank lines, found 0
vim-plug.py:26:21: E701 multiple statements on one line (colon)
vim-plug.py:28:1: E302 expected 2 blank lines, found 1
vim-plug.py:32:3: E301 expected 1 blank line, found 0
vim-plug.py:34:15: E711 comparison to None should be 'if cond is None:'
vim-plug.py:36:80: E501 line too long (86 > 79 characters)
vim-plug.py:41:3: E301 expected 1 blank line, found 0
vim-plug.py:50:1: E302 expected 2 blank lines, found 1
vim-plug.py:54:80: E501 line too long (81 > 79 characters)
vim-plug.py:77:80: E501 line too long (85 > 79 characters)
vim-plug.py:106:19: E261 at least two spaces before inline comment
vim-plug.py:116:1: E302 expected 2 blank lines, found 1
vim-plug.py:117:80: E501 line too long (83 > 79 characters)
vim-plug.py:127:7: E704 multiple statements on one line (def)
vim-plug.py:165:11: E128 continuation line under-indented for visual indent
vim-plug.py:188:15: E711 comparison to None should be 'if cond is not None:'
vim-plug.py:197:1: E302 expected 2 blank lines, found 1
vim-plug.py:219:7: E114 indentation is not a multiple of four (comment)
vim-plug.py:269:80: E501 line too long (80 > 79 characters)
vim-plug.py:278:1: E302 expected 2 blank lines, found 1
vim-plug.py:298:1: E302 expected 2 blank lines, found 1
vim-plug.py:313:1: E302 expected 2 blank lines, found 1
vim-plug.py:323:23: E226 missing whitespace around arithmetic operator
vim-plug.py:327:1: E302 expected 2 blank lines, found 1
```

Granted this is embedded code, but it's probably still good practice to follow PEP8.

plug.vim
```diff
@@ -968,6 +973,377 @@ while 1 " Without TCO, Vim stack is bound to explode
...
+class InvalidURI(Exception): pass
+class Action(object): INSTALL, UPDATE, ERROR, DONE = ['+', '*', 'x', '-']
+
+class GLog(object):
+  """ Global logger. Writes only when debug enabled. """
+  ON = None
+  LOGDIR = None
+  @classmethod
+  def write(cls, msg):
+    if cls.ON == None:
```
renstrom: if cls.ON is None

plug.vim
```diff
@@ -968,6 +973,377 @@ while 1 " Without TCO, Vim stack is bound to explode
...
+    self.lock = lock
+    self.maxy = int(vim.eval('winheight(".")'))
+    self.num_plugs = num_plugs
+
+  def where(self, name):
+    """ Helper function, given a name find first occurence in buffer lines. """
+    found, lnum = False, 0
+    matcher = re.compile('^[-+x*] {}:'.format(name))
+    for line in vim.current.buffer:
+      if matcher.search(line) != None:
```
renstrom: if matcher.search(line) is not None

plug.vim
```diff
@@ -968,6 +973,377 @@ while 1 " Without TCO, Vim stack is bound to explode
...
+          raise CmdTimedOut(['Timeout!'])
+
+      tfile.seek(0)
+      result = [line.rstrip() for line in tfile]
+
+      if pcmd.returncode != 0:
+        msg = ['CmdFailed:', self.cmd]
+        msg.extend(result)
+        raise CmdFailed(msg)
+    except:
```
renstrom: A bare `except` is probably not what you want. Change this to:

```python
except Exception:
```

>A bare `except:` clause will catch SystemExit and KeyboardInterrupt exceptions, making it harder to interrupt a program with Control-C, and can disguise other problems. If you want to catch all exceptions that signal program errors, use `except Exception:`

See https://www.python.org/dev/peps/pep-0008/ (under "Programming Recommendations") for more information about this.

plug.vim
```diff
@@ -968,6 +973,377 @@ while 1 " Without TCO, Vim stack is bound to explode
...
+      else:
+        self.install()
+        with self.lock:
+          vim.command("let s:update.new['{}'] = 1".format(self.name))
+    except (CmdTimedOut, CmdFailed, InvalidURI) as exc:
+      self.write(Action.ERROR, self.name, exc.message)
+    except KeyboardInterrupt as exc:
+      G_STOP.set()
+      self.write(Action.ERROR, self.name, ['Interrupted!'])
+    except:
```
renstrom: Same here.

plug.vim
```diff
@@ -968,6 +973,377 @@ while 1 " Without TCO, Vim stack is bound to explode
...
+    self.cmd_dir = cmd_dir
+    self.timeout = timeout
+    self.ntries = ntries
+
+    self.clean = clean
+    if cb:
+      self.callback = cb
+    else:
+      def dummy(msg): pass
+      self.callback = dummy
```
renstrom: Might as well use a `lambda` here:

```python
self.callback = lambda msg: None  # no-op
```

Writing `pass` is basically just another way of writing `return None` because all functions implicitly returns `None` if not otherwise specified:

```python
>>> def dummy(msg): pass
...
>>> foo = dummy('bar')
>>> type(foo)
<type 'NoneType'>
>>> print foo
None
```

renstrom: Or perhaps even reduce it a bit more (the parentheses can be omitted):

```python
self.callback = cb if cb else (lambda msg: None)
```

Or alternatively:

```python
if cb is None:
    cb = lambda msg: None  # no-op
self.callback = cb
```

starcraftman:
@renstrom Hello. All things considered, that isn't a terrible first run of PEP8. The `is None` is annoying, my brain just sometimes can't escape its C thinking. I do know better.

I don't mind making my code compliant, most of the spacing & double statement stuff I did to compress lines. Will review.

renstrom:
That's fine, I just thought I should bring it to light and then you guys can do what you will with it :)

plug.vim
starcraftman: I am aware of this, I'm kind of abusing broad `except:` lines to act as finally clauses that only execute on exceptional behaviour. Note, I am not suppressing the exceptions, I raise at the end. Just cleaning things up like my subprocesses, which should be cleaned up regardless what throws.

I wonder if python devs do this or use a boolean in a finally clause.

plug.vim
starcraftman: This I concur with, this is a perfect example of where a lambda is useful for saving a line. Thanks.

junegunn:
@starcraftman Looks good to me. Is there anything else you want to address?

plug.vim
renstrom: I see, fair enough.

renstrom:
I'm satisfied at least.

plug.vim
```diff
@@ -744,13 +745,17 @@ function! s:update_impl(pull, force, args) abort
   call append(0, ['', ''])
   normal! 2G
 
-  if s:ruby && s:update.threads > 1
+  if (s:py2|| s:ruby) && s:update.threads > 1
```
junegunn: Ah, one thing I forgot to mention.
Could you change this condition as follows, so that Neovim still prefers its own asynchronous installer?
```vim
if (s:py2 || s:ruby) && !s:nvim && s:update.threads > 1
```

starcraftman:
@junegunn Ya give me a bit of time, I'll ping you once I've finalized things and squashed down the commits again. Just want to do a sanity read on everything.

starcraftman:
@junegunn Just took out CmdFailed exception since it seems like in ruby code we just report timeout on anything not an interrupt. I believe its basically done, I have a feeling there may be a few more exception catching I'll have to add but people can raise tickets for that.

My only remaining complaint is that I poll, but short of a large refactor I don't see that changing.

starcraftman:
@vheon Do you mind doing a sanity check on this before it gets merged by exercising all the main paths? Just want to make sure I didn't mess anything up because of that recent fail.

starcraftman:
Note: I'll unavailable for most of the rest of today.

Want to do a bit more testing when I get back but should be mostly good.

vheon:
@starcraftman I'll do it tomorrow. Here is 00:09 am :P

starcraftman:
@vheon No problem, I'm in no rush. Sleep is important! Be sure to test timeout = 1, we had trouble with that prior. 

Barring some extreme case I haven't thought of, the threads/procs seem pretty well behaved on my Linux machine now.

vheon:
@starcraftman I've done some testing, all with `g:plug_timeout = 1` and seems to work well. You have to change the git commands based on the latest commit of vim-plug because the tag tracking has changed a bit.

starcraftman:
@vheon Done while you were testing. See last commit.

vheon:
@starcraftman oh ok :+1:

starcraftman:
@junegunn I think we are good to go. You may want to do one final review/test. I haven't run into any noticeable bugs/differences from ruby anymore. I also updated for #174, I believe that's the only recent commit that requires python changes.

@vheon Thanks for the testing.

junegunn:
Looks good to me! :shipit: 
Thank you very much for your hard work!

starcraftman:
@junegunn No problem, was an interesting project.

Issue #161, opened 2015-02-08
`GIT_TERMINAL_PROMPT=0` git 2.3 feature


bruno-:
Hi,
thank you for making and maintaining this great plugin.

A couple days ago new git version `2.3` was released. Among new features is also `GIT_TERMINAL_PROMPT` environment variable. When it is set to `0` git will never prompt for user credentials.
You can learn more about it in `man git` (after updating to git 2.3) or in [this blog post on github](https://github.com/blog/1957-git-2-3-has-been-released).

Just as an example, here's how it works:
- `git clone https://github.com/some/non-existing-repo` - will prompt for a username & password
- `GIT_TERMINAL_PROMPT=0 git clone https://github.com/some/non-existing-repo` - will fail without prompting for username & password
- `GIT_TERMINAL_PROMPT=0 git submodule add https://github.com/chaconinc/DbConnector` - also fails withot prompting for username & pass

I think this addition could potentially be valuable to `vim-plug` in the following places:
- when doing `git clone`
- when doing `git submodule update`
- when doing `git fetch` (just in case a user manually changed git remote)

I was thinking of opening a pull request that prepends git commands with `export GIT_TERMINAL_PROMPT=0`. Example line 926:

```bash
(export GIT_TERMINAL_PROMPT=0; git fetch --progress 2>&1 && git checkout -q %s 2>&1 && git merge --ff-only origin/%s 2>&1 && git submodule update --init --recursive 2>&1)
```

I did something similar for [tmux plugin manager](https://github.com/tmux-plugins/tpm/commit/9949d48b6afa105c9e8489324b74789a92270a9a).

Related issues are #109, #56.

What do you think about this?

junegunn:
Thanks for the info! I wish this was available in the older versions of git. :disappointed: 
I'll definitely consider applying this new option and (finally) removing the annoying `git::` from the url, but since we can't expect the majority of the users of vim-plug to upgrade their git to 2.3 anytime soon, I'll leave this issue open for a while if you don't mind. We also have to check how we can apply this on Windows.

bruno-:
Hi @junegunn,
no problem about leaving this open.

About the `git::` - of course that should be left in the source for backwards compatibility.

[...]

PR #168, opened 2015-02-15
Do not inject `git::` to url when git 2.3.0 or above found

junegunn:
Related: #161 
/cc @bruno-

This commit takes advantage of `$GIT_TERMINAL_PROMPT` variable added in git 2.3.0. So if the git installed on the system is 2.3.0 or above, vim-plug will not use `git::` hack to prevent user prompt. However, the downside of the current implementation is that it runs `git --version` command on startup which should increase the startup time by a little.

### DIFF
--- a/plug.vim
+++ b/plug.vim
@@ -718,6 +718,15 @@ function! s:update_impl(pull, force, args) abort
     return
   endif
 
+  if !s:is_win && s:git_version_requirement(2, 3)
+    let git_terminal_prompt = exists('$GIT_TERMINAL_PROMPT') ? $GIT_TERMINAL_PROMPT : ''
+    let $GIT_TERMINAL_PROMPT = 0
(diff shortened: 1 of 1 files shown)


junegunn:
There are two alternatives that come to mind:
1. Defer setting up URI until it's needed. However, this requires some changes in `g:plugs` structure.
2. Remove `git::` part from uri using string substitution before update/install.

junegunn:
Implemented the second method. (We may need to revert the value of GIT_TERMINAL_PROMPT though.)

bruno-:
Hi @junegunn,
:+1: 
Much more elegant code than if I had submitted a pull request.
Also, I updated `vim-plug` locally and tested it a bit and things seem to work fine.

junegunn:
Thanks. I appreciate your feedback!

PR #179, opened 2015-02-22
Linking to wiki.

starcraftman:
Just putting the link now on the front page. Wiki is up, mainly just notes on Ubuntu. Compiling should cover any generic *NIX machine. And Mac I'm told has support out of box so just left it unsaid.


junegunn:
Thanks.

PR #187, opened 2015-03-07
Python 2.6 compatibility

junegunn:
Tested on:
- Vim 7.2.330
- Python 2.6.5
- Ubuntu 10.04
- Git 1.7.0.4 / 2.3.2
- https://gist.github.com/junegunn/863dcbd78eadece788cd

/cc @starcraftman @vheon 

I'm not completely sure if there's anything else to be done, but it seems like everything works as expected and the required change was kind of trivial. I'll merge after some additional tests.

### DIFF
--- a/plug.vim
+++ b/plug.vim
@@ -764,7 +764,7 @@ function! s:update_impl(pull, force, args) abort
     silent python import platform; print(platform.python_version())
     redir END
     let s:py2 = s:version_requirement(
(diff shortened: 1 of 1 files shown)


vheon:
Seems good to me. That was exactly the changes that I was expecting.

starcraftman:
@junegunn Looks fine to me, I think the default travis image is still python 2.6, so you can have a separate test case for it or simply remove the install of 2.7. Though the code is functionally identical so might be a bit moot.

Issue #188, opened 2015-03-12
Updating fails even with Python 2.6+ installed


cparadis6191:
I've been using vim-plug for a while with no problems but recently `:PlugUpdate` started to fail with

```bash
Updated. Elapsed time: 2.255018 sec.
[xx==]

- Finishing ... Done!
- vim-fugitive: Already up-to-date.
- rainbow_parentheses.vim: Already up-to-date.
x vimwiki
    [87;1H[?1l>[?12l[?25h[?1049lVim: Caught deadly signal SEGV
x syntastic
    [87;1H[m[?1l>[?12l[?25h[?1049lVim: Caught deadly signal SEGV
```

This happens on a default Cygwin install (with vim and git installed) with the default Python installation.

```bash
$ python --version
Python 2.7.8
```

```bash
$ vim --version
+python/dyn
+python3/dyn
+ruby/dyn
```

Installing Ruby fixes the problem and allows normal updating.

junegunn:
Hmm, segmentation fault. This is probably due to the defect of the python interface of Vim you're running. Could you post the output of `:python import sys; print sys.version` from Vim?

/cc @starcraftman

cparadis6191:
Sure thing. That gives

```bash
2.7.8 (default, Jul 28 2014, 01:34:03)
[GCC 4.8.3]
```

I'll try a different version of python sometime tomorrow when I get a chance and see if the problem persists.

junegunn:
Thanks, I thought it was a similar problem to the one described [here](https://github.com/junegunn/vim-plug/wiki/faq#vim-caught-deadly-signal-segv). But the problem seems to be more complicated than I imagined.

> I've been using vim-plug for a while with no problems but recently :PlugUpdate started to fail 

Did vim-plug installer work in parallel back then? Or did it install plugins one by one?

starcraftman:
@junegunn Interesting! Firstly, I hadn't really done any testing on cygwin, though I hadn't ruled out its use as it is technically posix, os.name reports posix under cygwin. So that's why he's using parallel even though I ruled out Windows support. 

I just ran a simple test with my own vimrc under Babun (a Cygwin wrapper) and found nicely that vim-plug install/update with python does indeed work under cygwin. Interrupting even works, though it is slow. Didn't see any issue. Though oddly, my GLog logger doesn't seem to work, odd since it is so simple. :(

[...]

PR #189, opened 2015-03-12
Disable Python on Cygwin

starcraftman:
Temporary fix for #188, need to further investigate to know how broken cygwin support is.


junegunn:
Thanks for the quick response. It would be nice if we could workaround the problem, but man it's nasty segmentation fault, and probably there's nothing we can do about it.

starcraftman:
@junegunn I agree it is quite nasty. It appears to be a concurrency issue as nthreads = 1 leads to no problem. For some reason, the threading lock isn't preventing multiple threads from updating the buffer & calling vim commands. I assume this may be a cygwin bug as I've never had such an issue on my posix machines.

junegunn:
Yeah, probably. Don't try too hard. I'm on OS X and Linux 99% of time and it's really hard for me to find time to deal with Windows issues (purely for somebody else).

Issue #81, opened 2014-09-17
Parallel installer on Windows GVim does not show progress


saaguero:
Could be this because of this issue? https://github.com/junegunn/vim-plug/issues/4

I mean, it works perfectly, but I don't have any feedback about the progress except for many consoles being opened, I supposed that each one represents a git process cloning the corresponding plugin.

If that is the case, have you tried any solution? Or something that's blocking the implementation?

starcraftman:
@junegunn I ran into this too, annoyingly. I reported it to the vim_dev list (1) because the problem seems to lie directly with GVim rather than something in this plugins code. Seems that with the Windows version, when you are inside an embedded language you can't call redraw. Ever. I suspect the main UI loop may be blocked waiting for command to finish? 

Due to this, I'm not certain what I'm going to do for Windows. Debating whether to just leave it disabled like ruby since its so poor to see GVim freeze until command unblocks.

1) https://groups.google.com/forum/#!topic/vim_dev/Pmisv__GAFo

junegunn:
@starcraftman Yeah, I've been considering disabling parallel installer on Windows altogehter because the experience is really horrible and unstable. It's perfectly fine for me if you decide to give up supporting Windows on your python implementation.

PR #191, opened 2015-03-14
Disable Ruby parallel installer on Windows.

vaibhavsagar:
References #81

### DIFF
--- a/plug.vim
+++ b/plug.vim
@@ -74,7 +74,7 @@ let s:plug_buf = get(s:, 'plug_buf', -1)
 let s:mac_gui = has('gui_macvim') && has('gui_running')
 let s:is_win = has('win32') || has('win64')
 let s:py2 = has('python') && !s:is_win && !has('win32unix')
-let s:ruby = has('ruby') && (v:version >= 703 || v:version == 702 && has('patch374'))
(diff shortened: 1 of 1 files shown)


vaibhavsagar:
I'll finally be able to remove this snippet from my `.vimrc`, and you'll be able to remove it from the wiki:
```vim
if (has('win32') || has('win64'))
    let g:plug_threads=1
endif
```

junegunn:
The Ruby installer on Windows does work even though the experience is suboptimal. We may decide to make Ruby installer an "opt-in" feature (e.g. only when `g:plug_threads` is set, or when the number of threads is explicitly specified; `:PlugInstall 8`) but I don't want to permanently disable it. Also, there are conditional branches in the Ruby code for Windows, so even if I'm going to disable it altogether, I'm not going to do it without properly cleaning up the code.

I'll let you know of the progress. Thanks.

vaibhavsagar:
That's definitely a better solution, thanks for the quick response!

junegunn:
So here it goes, now the default number of threads on Windows is 1. I also updated the wiki page. Thanks for your suggestion.

vaibhavsagar:
Thank you!

Issue #193, opened 2015-03-21
Missing status when PlugUpdate in Neovim 


vheon:
In neovim I get:
```
Updated. Elapsed time: 7.766691 sec.
[=========================================]

- Finishing ... Done!
- Post-update hook for vim-go ... Done!
- vim-argumentative: 
- vim-pandoc-syntax: 
- YouCompleteMe: 
- vim-scriptease: 
- vim-lion: 
- vim-colors-solarized: 
- vim-repeat: 
- vim-tmux: 
- vimerl: 
- ultisnips: 
- vim-pandoc: 
- vim-fish: 
- vim-go: 
- linediff.vim: 
- splitjoin.vim: 
- vim-commentary: 
- vim-java: 
- goyo.vim: 
- vim-rsi: 
- rust.vim: 
- vim-abolish: 
- MatchTagAlways: 
- vim-help-grimoire: 
- vim-fugitive: 
- LargeFile: 
- vim-sleuth: 
- vim-surround: 
- cpp-vim: 
- vim-visual-star-search: 
- vim-sensible: 
- vim-exchange: 
- ListToggle: 
- vim-dirvish: 
- vim-unimpaired: 
- inline_edit.vim: 
- tagbar: 
- vim-textobj-delimited: 
- ctrlp.vim: 
- vim-eclim: 
- lexima.vim: 
- vim-cpp-enhanced-highlight: 
```

while in vim I get:
```
Updated. Elapsed time: 8.328021 sec.
[=========================================]

- Finishing ... Done!
- vim-fish: Already up-to-date.
- vim-pandoc-syntax: Already up-to-date.
- ultisnips: Already up-to-date.
- vim-go: Already up-to-date.
- vim-argumentative: Already up-to-date.
- YouCompleteMe: Already up-to-date.
- vim-lion: Already up-to-date.
- vim-repeat: Already up-to-date.
- vim-pandoc: Already up-to-date.
- vim-scriptease: Already up-to-date.
- vimerl: Already up-to-date.
- vim-tmux: Already up-to-date.
- vim-colors-solarized: Already up-to-date.
- command-t: Already up-to-date.
- linediff.vim: Already up-to-date.
- splitjoin.vim: Already up-to-date.
- vim-commentary: Already up-to-date.
- vim-java: Already up-to-date.
- goyo.vim: Already up-to-date.
- vim-rsi: Already up-to-date.
- rust.vim: Already up-to-date.
- vim-abolish: Already up-to-date.
[...]

PR #194, opened 2015-03-21
Fix for #193 Missing status when PlugUpdate

starcraftman:
Seems to fix the problem for me. I do not know how long this was an issue though.

@vheon All good for you?

Edit: Work in progress, don't actually pull until I resolve all this.

### DIFF
--- a/plug.vim
+++ b/plug.vim
@@ -849,7 +849,7 @@ function! s:job_handler(name) abort
     call s:reap(a:name)
     call s:tick()
   else
-    let job.result .= s:to_s(v:job_data[2])
+    let job.result .= substitute(s:to_s(v:job_data[2]), '[\r\n]', '', 'g') . "\n"
     " To reduce the number of buffer updates
     let job.tick = get(job, 'tick', -1) + 1
     if job.tick % len(s:jobs) == 0


vheon:
Now the status is shown, but when there are some changes the statuses are different, but I don't really know how was before.

In vim:
`- vim-lion:  1 file changed, 20 insertions(+), 4 deletions(-)`

In neovim:
`- vim-lion: Updating 5fa766e..903d4ccFast-forward plugin/lion.vim | 24 ++++++++++++++++++++---- 1 file changed, 20 insertions(+), 4 deletions(-)
`

starcraftman:
@vheon Hmmm, doesn't really help that neither of us know what it did before... I just checked and it appears the culprit was this line  `  852     let job.result ,= s:to_s(v:job_data[2])`, we were concatenating the results rather than replacing. Now shows live update. Though, oddly, it ends with just done instead of connectivity..... done.

junegunn:
@starcraftman `job.result` holds the entire output of the process, so it is appended. The reason we do this is to show the full message on error. See `s:format_message`.

@vheon I'd like to know why and when the output of neovim was changed. Maybe I can install some older versions of neovim to find out.

starcraftman:
@junegunn Ah I see, very well. I changed my branch to preserve the entire output in job.result & simply added job.last_line to track last line to output under normal conditions. Changed s:reap to not clobber buffer with whole job.result. I think this more closely mirrors behaviour of ruby/py live displaying the last line.

starcraftman:
Also wow, travis seems to either be overloaded or just de prioritizing free people. Committed 20 mins ago & not one build has run.

plug.vim
```diff
@@ -935,7 +940,8 @@ function! s:log(bullet, name, lines)
     else
       let pos = 4
     endif
-    call append(pos - 1, s:format_message(a:bullet, a:name, a:lines))
+    let filtered = substitute(a:lines, '[[:cntrl:]]', '', 'g')
```
junegunn: So what are the `[[:cntrl:]]` characters we are filtering here? Is this an issue of Neovim? And if so, do you think we should do this as a safety net even when the issue is resolved?

starcraftman: 1) When I dump out the contents of a:lines before filtered what I see is most lines end in `\r\n`, while the first few oddly end in `\n\n`. I'm not sure why we are getting Windows EOL when I'm on Linux. So the [[:cntrl:]] could be replaced with [\r\n] if we want to be exact. 

2) Since I don't know what neovim was doing before, can't comment yet. Did older versions only have 1 newline? Trimmed? Perhaps in a recent commit they changed the EOL for some other feature. Might build some older commits to see.

3) Well as a safety issue we could leave it, when using append you aren't ever supposed to leave any new line control characters in the buffer. Got caught by that on Python a few times.

Also this new EOL policy broke s:lines. Quite a nuisance.

starcraftman:
@junegunn I just did the bisect, should have done to start with. Commit that breaks our code follows. Apparently, we've been broken for a month.

```
* d28011e (Mon Feb 23 19:43:33 2015) [Thiago de Arruda] eval: Fix buffering of data in job autocommands
| Job autocommands will no longer buffer data chunks that don't end in newlines
| characters.
| 
| 48    50      src/nvim/eval.c
| 25    8       test/functional/job/job_spec.lua
```

starcraftman:
@junegunn Related question, why are we appending job.result instead of just storing a list of strings? In the latter case then we can just join when we want more than last. Would immunize against such problems & ease EOL stripping. I need food, will refactor my branch to that I think later.

junegunn:
@starcraftman Because neovim installer was written based on the original vimscript implementation where the whole output is obtained using a single `s:system()` call. As long as we keep both versions to share as much code as possible and not deviate much from each other, I'm not against changing it to a list of strings.

starcraftman:
@junegunn Alright, I'll see what's easiest to do and ping you when I've got something on my branch worth a look.

junegunn:
@starcraftman Thanks! By the way, you want to fix it before neovim changes its API?

starcraftman:
@junegunn Condensed it down into a one line fix. Works for me with latest neovim. If you want to you can pull it now. The more major change will be separate, should open a new issue and paste in tarruda's link.

junegunn:
Thanks!

PR #196, opened 2015-03-23
Neovim Interrupt Jobs

starcraftman:
When I was testing nvim, I noticed a lack of interrupt support like our other two parallel methods. This works for me, just a small hack.


starcraftman:
Need to review, just realized need to go through finish functions too. Will repost in a bit.

junegunn:
Neovim installer is asynchronous, and you can be on any buffer while the installation is ongoing. So the traditional way of interrupting tasks with CTRL-C does not make a lot of sense. Instead, closing the vim-plug buffer will abort the tasks.

```vim
  if !s:plug_window_exists() " plug window closed
    return s:job_abort()
  endif
```

starcraftman:
@junegunn Ah I see, oversight on my part. Makes sense. I guess I'm just so used to blocking on commands in vim that I was used to seeing all those interupted messages.....

PR #198, opened 2015-03-26
New Job API

starcraftman:
Make us work with new api. To be merged after neovim merges PR neovim/neovim#2247.

Notes:
* Disabled python & ruby on nvim. The current implementations won't work. Not sure if there's any need, since job support in nvim is mandatory.
* Jettisoned the echo Error trick in favor of just checking return code on job exit.

### DIFF
--- a/plug.vim
+++ b/plug.vim
@@ -72,9 +72,9 @@ let s:plug_tab = get(s:, 'plug_tab', -1)
 let s:plug_buf = get(s:, 'plug_buf', -1)
 let s:mac_gui = has('gui_macvim') && has('gui_running')
 let s:is_win = has('win32') || has('win64')
-let s:py2 = has('python') && !s:is_win && !has('win32unix')
-let s:ruby = has('ruby') && (v:version >= 703 || v:version == 702 && has('patch374'))
-let s:nvim = has('nvim') && !s:is_win
+let s:py2 = has('python') && !has('nvim') && !s:is_win && !has('win32unix')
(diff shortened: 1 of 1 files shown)


junegunn:
Great job! I'll merge when it's ready.

plug.vim
```diff
@@ -863,18 +862,22 @@ function! s:spawn(name, cmd, opts)
...
-            \ (has_key(a:opts, 'dir') ? s:with_cd(a:cmd, a:opts.dir) : a:cmd)
-            \ . ' || echo Error'])
-    if x > 0
-      let job.jobid = x
-      augroup PlugJobControl
-        execute 'autocmd JobActivity' a:name printf('call s:job_handler(%s)', string(a:name))
-      augroup END
+    let opts = {
+        \ 'on_stdout': function('s:job_handler'),
+        \ 'on_exit' : function('s:job_handler'),
```
junegunn: Just a minor nitpicking. There's a space before colon.

starcraftman: Those sneaky spaces!

tarruda:
@starcraftman there was a small change to the API("user" option was removed and the job options are passed as the "self" argument)

starcraftman:
@tarruda Done. Seems like an improvement, easy to have user object simply contain the callbacks required.

starcraftman:
@junegunn Ping.  You can pull now. Neovim PR went through, I don't see any drastic changes to API from my last update. Ran a test against neovim & Install/Update works fine.

junegunn:
:tada:

PR #201, opened 2015-04-02
PlugUpgrade using git (close #200)

### LINKED ISSUE #200: Error messages of PlugUpgrade with Python is not helpful
I have compiled my own Vim with user compiled Python 2.7, and when I do `:PlugUpgrade`, it says:
```
Error upgrading vim-plug: Vim(python):Traceback (most recent call last):
```

But I can't see the detailed error messages even if I call `:mess`. It just shows that 'traceback' line. I think the cause of this error of `:PlugUpgrade` is not related to this plugin, but it should give me detailed traceback.

junegunn:
(#200)

@starcraftman @yous

So here's the new upgrade code using git. What do you think?
I'm not 100% sure if it would run on Windows :confused:

### DIFF
--- a/plug.vim
+++ b/plug.vim
@@ -67,7 +67,7 @@ let g:loaded_plug = 1
 let s:cpo_save = &cpo
 set cpo&vim
 
-let s:plug_src = 'https://raw.githubusercontent.com/junegunn/vim-plug/master/plug.vim'
+let s:plug_src = 'https://github.com/junegunn/vim-plug.git'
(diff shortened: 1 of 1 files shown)


yous:
At least this works with [Yongwei's gVim](http://wyw.dcweb.cn/#download) with patches 1-589 are included.

junegunn:
@yous Thanks for testing it! :+1: I guess then we can merge it.

PR #208, opened 2015-04-08
Testing changes.

starcraftman:
Two things here:
1. Vim project moving to github eventually on a permanent basis as code.google is closing. For now I believe it is a synchronized mirror while they learn/debug migrating. Changing will future proof us.
2. Took liberty of using depth=1, no problems & build time cut in half. Wish hg supported that.


junegunn:
Good stuff @starcraftman. Thanks!

PR #211, opened 2015-04-12
Ignore case for "<Plug>"

mhinz:
안녕하세요!

This is really just personal style, but I prefer my keys to be lowercase, hence this minimal change. 

Actually I wondered why it wouldn't autoload when I did `... { 'on': '<plug>(func)' }`. Could be unexpected for other users, too. :)


junegunn:
Danke schön!

PR #219, opened 2015-04-27
[RFC] Support Python 2.6+ & 3.0+

starcraftman:
* Minimal python code changes required.
* Now write temp file then use py(3)file.

Temp file is generated once on invoking update_python, then regenerated only on sucessful PlugUpgrade. I've tested on 3.1.2 & 3.4.0. Changes do not seem to have any impact on 2.6+ support from what I've seen.

@junegunn Do you know why my py3 (2.1.0) test run fails on that one test?

### DIFF
--- a/plug.vim
+++ b/plug.vim
@@ -72,7 +72,8 @@ let s:plug_tab = get(s:, 'plug_tab', -1)
 let s:plug_buf = get(s:, 'plug_buf', -1)
 let s:mac_gui = has('gui_macvim') && has('gui_running')
 let s:is_win = has('win32') || has('win64')
-let s:py2 = has('python') && !has('nvim') && !s:is_win && !has('win32unix')
+let s:py = (has('python') || has('python3')) && !has('nvim') && !s:is_win && !has('win32unix')
+let s:py_exe = has('python3') ? 'python3' : 'python'
 let s:ruby = has('ruby') && !has('nvim') && (v:version >= 703 || v:version == 702 && has('patch374'))
 let s:nvim = has('nvim') && exists('*jobwait') && !s:is_win
 let s:me = resolve(expand('<sfile>:p'))
@@ -754,7 +755,7 @@ function! s:update_impl(pull, force, args) abort
     \ 'pull':    a:pull,
     \ 'force':   a:force,
     \ 'new':     {},
-    \ 'threads': (s:py2 || s:ruby || s:nvim) ? min([len(todo), threads]) : 1,
+    \ 'threads': (s:py || s:ruby || s:nvim) ? min([len(todo), threads]) : 1,
     \ 'bar':     '',
     \ 'fin':     0
   \ }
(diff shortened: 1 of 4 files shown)


vheon:
> Now write temp file then use py(3)file.

Can I ask why?

plug.vim
```diff
@@ -996,12 +997,26 @@ endwhile
...
 function! s:update_python()
-python << EOF
+  if !filereadable(s:py_script)
+    let flist = readfile(s:me)
+    let start = index(flist, "python << PYEOF") + 1
+    let end = index(flist, "PYEOF") - 1
+    call writefile(flist[start : end], s:py_script)
+  endif
+  let py_prog = has('python3') ? 'py3file' : 'pyfile'
+  execute py_prog . ' ' . s:py_script
```
vheon: You can change this line to `execute py_prog s:py_script`. In the rest of the code is done this way when possible.

starcraftman:
@vheon Well, from what I saw, you can't use the heredoc syntax  with a variable (i.e. the line I do python << EOF). So to support py3 I'd have to copy the whole code a second time in a python3 << EOF section. If I am mistaken, I'm open to an alternative.

Clearly we could duplicate like that in the file, I just don't want to.

vheon:
The code executed is the same if I use python2 or python3 right? I don't see the problem in leaving the code in the function `s:update_python` and call it like before and use "our" python like:


```viml
let s:ourpy = has('python3') ? 'python3' : 'python'
```

and in `s:update_python`

```viml
function! s:py_nop()
execute s:outpy "<< PYEOF"
```

starcraftman:
@vheon Weird, could have sworn that didn't work, maybe I typod :/. Lemme go make that change. Much better.

junegunn:
> Do you know why my py3 (2.1.0) test run fails on that one test?

Because py3 version prints `+ aaaa...` instead of `x aaaa...`. Compare the following logs:

```
    ( 59/102) [EXECUTE] Retry failed tasks
      > ['Updated. Elapsed time: 0.087583 sec.', '[=x]', '', '- Finishing ... Done!', '- vim-easy-align: Already installed', 'x aaaaaaaaaaaaaa', '    Cloning into ''/tmp/vHhc565/6/aaaaaaaaaaaaaa''...', '    fatal: ''/tmp/junegunn/aaaaaaaaaaaaaa'' does not appear to be a git repository', '    fatal: Could not read from remote repository.', '    ', '    Please make sure you have the correct access rights', '    and the repository exists.']
```
```
    ( 59/102) [EXECUTE] Retry failed tasks
      > ['Updated. Elapsed time: 0.273976 sec.', '[= ]', '', '- Finishing ... Done!', '- vim-easy-align: Already installed', '+ aaaaaaaaaaaaaa: and the repository exists.']
    ( 59/102) [EXECUTE] (X) 0 should be equal to 1
```

starcraftman:
@junegunn Ya it just kinda seems odd since same code. Will trace tomorrow. Otherwise, will be ready to go once fixed.

junegunn:
@starcraftman I tested locally on OS X, and got the same error as on travis.

starcraftman:
@junegunn Ya I see it locally now. I'll get to that. 

Regarding vader, is there a simple way to interrupt the flow by breakpoint so you can see the current state of vim. Say in the case of this problem, at retry test case on the fail. So far I resort to editting the test file to excludes tests I don't want.

junegunn:
@starcraftman No, it would be nice though.

starcraftman:
@junegunn Found problem, seems py3 changed exception objects. I overlooked the possibility that they'd deprecate the message property, which is what they did. Should be good to go. 

Also, sorry for delay, was busy past few days.

junegunn:
Thanks!

Issue #217, opened 2015-04-24
add option to only use HEAD version of git when installing


marcmo:
vim-plug is really fast when you update your plugins. but what can take a rather long time to do a fresh install (some plugins have a rather long git history).
but actually I'm not interested in the history.

e.g. I just installed YCM which is a rather huge repo:

Updated. Elapsed time: 170.097284 sec.
[============================================]
...

191M    ./YouCompleteMe

if I just clone the HEAD version it with git:

```
> time git clone --depth=1 https://github.com/Valloric/YouCompleteMe.git
Cloning into 'YouCompleteMe'...
remote: Counting objects: 72, done.
remote: Compressing objects: 100% (67/67), done.
remote: Total 72 (delta 3), reused 41 (delta 0), pack-reused 0
Unpacking objects: 100% (72/72), done.
Checking connectivity... done.

real    0m2.319s
user    0m0.090s
sys 0m0.084s
```

compare that to the full repo:

```
> time git clone https://github.com/Valloric/YouCompleteMe.git
Cloning into 'YouCompleteMe'...
remote: Counting objects: 29994, done.
remote: Compressing objects: 100% (7/7), done.
remote: Total 29994 (delta 0), reused 0 (delta 0), pack-reused 29987
Receiving objects: 100% (29994/29994), 29.15 MiB | 576.00 KiB/s, done.
Resolving deltas: 100% (9690/9690), done.
Checking connectivity... done.

real    1m0.539s
user    0m2.410s
sys 0m2.568s
```

vheon:
Duplicate of #112?

junegunn:
Yep shallow clone and sparse checkout were suggested in #122. I still don't see the point of sparse checkout, but now I think shallow clone can be a good addition to vim-plug. I was reluctant at first because of some issue reports, for example this one: https://github.com/Shougo/neobundle.vim/issues/81#issuecomment-17161962. But it looks like the problem is in `git fetch --depth 10` and not in `git clone --depth 1`. I don't think vim-plug will ever have to `fetch` plugin with `--depth` option, so it's not something we have to worry about. And I personally had no issue with shallow clones. What do you guys think? If we can be sure that shallow clone does not cause any problem, we can probably do it as the default, instead of adding an option. Also, in order to support switching of branches we're going to put `--no-single-branch` option as weel.

starcraftman:
[...]

Issue #122, opened 2014-11-21
Reduce data it downloads


Tranquility:
You can use sparse checkouts (to download only the directory specified in "rtp") and shallow clone (to not download the full history) to drastically reduce the data it needs to download.

This is especially important with big repositories like rust. For exacmple: 

```viml
 Plug 'rust-lang/rust', {'rtp': 'src/etc/vim'}
```

junegunn:
Closing by #218.

junegunn:
The patch was reverted. The discussion will continue on #217.

PR #221, opened 2015-04-28
Shallow clone by default (#122 #217)

junegunn:


junegunn:
`--unshallow` should be used only once. :astonished:

junegunn:
And `fetch --unshallow` requires git 1.8.3 or above.

junegunn:
And `git fetch --depth 99999999` is slower than `git fetch`.

junegunn:
So here's the summary of this PR:

1. Shallow clone (`--depth 1`)
    1. if `g:plug_shallow` is not set to 0 and `tag` is not specified.
    1. `--no-single-branch` option is also used for `branch` option
1. `PlugUpdate` will `git fetch --depth 99999999` when `tag` option is given to "unshallow" the repository
    1. Not `--unshallow` because it requires git 1.8.2 or above
    1. `--depth 99999999` is attached *only when* the repository is indeed shallow (`.git/shallow` exists) because `depth` makes fetch noticeably slower.

PR #226, opened 2015-05-04
Allow custom repo-to-name modifier

myfreeweb:
Use case: remove the ':t' from the modifier, call plug#begin with '~/src/github.com' and keep the repos in a `go get` / [`ghq`](https://github.com/motemen/ghq) style workspace.

### DIFF
--- a/plug.vim
+++ b/plug.vim
@@ -424,7 +424,7 @@ function! s:add(repo, ...)
 
   try
     let repo = s:trim(a:repo)
-    let name = fnamemodify(repo, ':t:s?\.git$??')
+    let name = fnamemodify(repo, get(g:, 'plug_name_modifier', ':t:s?\.git$??'))
     let spec = extend(s:infer_properties(name, repo),
                     \ a:0 == 1 ? s:parse_options(a:1) : s:base_spec)
     if !has_key(g:plugs, name)


junegunn:
I'm not particularly against different way of organizing stuff, but I'm not convinced that it's needed either. Could you elaborate on the benefit of such directory structure? I'm familiar with GOPATH and stuff, but I don't see why one would like that for vim plugins. Switching plugins with the same name without having to remove the current one could be one reason I guess, but it rarely happens in practice and doesn't seem like a good reason to add another option that most users wouldn't care or need.

And even if we decided to support different structure, I don't think your approach is optimal for the following reasons:
1. Usability consideration: I'm pretty certain that most users will not be able to understand how to configure the variable. Too low-level in my opinion.
2. Doesn't work nicely with raw git urls. e.g. `Plug 'https://github.com/junegunn/goyo.vim.git'` or `Plug 'git@github.com:junegunn/goyo.vim.git'`
3. Not quite like `go get` / `ghq` where the host name of a url becomes a part of the path. With vim-plug you can have plugins not only from GitHub, but also from Bitbucket or any other git server. `go get` would separate them with their host names, but your approach does not.

junegunn:
I forgot to mention that a git url can point to a path on the file system.

```vim
" Manually installed/updated. vim-plug is only responsible for managing &runtimepath
Plug '~/foobar'

" Cloned/pulled just like any other git urls
Plug 'file:///root/foobar'
```

myfreeweb:
Maybe I shouldn't have written about my use case... :D

It works *for me*. I only use github repo names and nothing else (so `github.com` is the root folder). I don't want this to be documented anywhere.

Having a proper algorithm that separates the hostname and the path would be very cool, but looks like you don't want it.

I don't know any "real" benefits, I just *want* all git repos on my machine to always be in that directory structure.

junegunn:
> It works for me. I only use github repo names and nothing else (so github.com is the root folder).

That's fine. But as the maintainer of the project, I have to make sure that it works for *all cases* that vim-plug supports. You should probably use your fork.

> Having a proper algorithm that separates the hostname and the path would be very cool, but looks like you don't want it.

I don't know if I want it or not, the benefits are just unclear to me. I was more saying that if we're going to do it, we have to do it right, I don't want a half-baked solution. Anyway, thanks for the suggestion.

Issue #104, opened 2014-10-10
Neovim installer should be synchronous during `vim_starting`


junegunn:
Since the new parallel installer for Neovim is asynchronous, we cannot do `nvim +PlugUpdate +qa` as before. We can either add an option to `PlugUpdate` command to quit Vim when it's complete or wait for the addition of `jobwait()` function suggested by @tarruda. Or we can simply guide the users to use the original Vim instead in this case.
See:
- https://github.com/junegunn/vim-plug/pull/103#issue-45353388
- https://github.com/junegunn/vim-plug/pull/103#issuecomment-58603245
    - https://github.com/junegunn/vim-plug/pull/103#issuecomment-58605181

adambiggs:
I'm trying to make a Docker image for my dev environment, but I hit the wall at this point. 

`RUN vim +PlugInstall +qa` starts installing the first dozen or so plugins, but then quits right away before completion. Then when I run the docker image, open `nvim`, and run `:PlugInstall` again, the plugins that started installing previously are listed with the following error:

```
HEAD
PlugClean required.
```

I tried using `nvim +'PlugInstall --quit'` like you mentioned in #103, but it looks like this isn't implemented yet. Is there any workarounds you know of? Doesn't matter if it's hacky since it only gets run once when building the docker image.

Thanks again for the great tool!

junegunn:
@adambiggs 

> RUN vim +PlugInstall +qa in my Dockerfile starts installing the first dozen or so plugins, but then quits right away before completion

This shouldn't happen. I also have my Dockerfile which does `PlugInstall`, and it installs just fine. Make sure that you're running the original vim here, not neovim. Is it possible that `vim` is somehow linked to `nvim`? Check out the output of `vim --version`.

adambiggs:
@junegunn thanks for the quick reply. I'm running `nvim` on purpose in this docker image. But I guess I can just add normal `vim` to the image and use it to install plugins in the Dockerfile.

Probably a good idea to have both versions anyway until NeoVim becomes more stable.

Thanks for your help!

tarruda:
It will be possible to close this after neovim/neovim#2247 is merged.

tarruda:
I've been thinking of how [jobwait()](http://neovim.org/doc/user/eval.html#jobwait()) can be used to implement this.

[...]

PR #227, opened 2015-05-05
[RDY] Neovim Python & Synchronous +PlugInstall

starcraftman:
* Buffer updates now managed by `__main__` loop with buf_q. Entirely for neovim support.
* Synchronous neovim install temporarily provided by python installer.

Known issues:
*   No ctrl-c/interrupt support on nvim.
*   Graphical bug: neovim/python-client#103
*   ~~Travis Python 3 build segfaults.~~

@junegunn Was interested in this to respond to #104 . This PR is work in progress still need to fix segfault. Open to comments.

Open Questions:
* Readme changes? Should it be noted in the front page like: **ATTENTION NEOVIM USERS** for sync install etc...
* Should nvim +PlugInstall **require** the neovim python lib. That is to say, don't do any installation from starting to prevent asynchronous install entirely or fallback to neovim's installer?

### DIFF
--- a/plug.vim
+++ b/plug.vim
@@ -72,9 +72,6 @@ let s:plug_tab = get(s:, 'plug_tab', -1)
 let s:plug_buf = get(s:, 'plug_buf', -1)
 let s:mac_gui = has('gui_macvim') && has('gui_running')
 let s:is_win = has('win32') || has('win64')
-let s:py = (has('python') || has('python3')) && !has('nvim') && !s:is_win && !has('win32unix')
-let s:py_exe = has('python3') ? 'python3' : 'python'
-let s:ruby = has('ruby') && !has('nvim') && (v:version >= 703 || v:version == 702 && has('patch374'))
 let s:nvim = has('nvim') && exists('*jobwait') && !s:is_win
 let s:me = resolve(expand('<sfile>:p'))
 let s:base_spec = { 'branch': 'master', 'frozen': 0 }
@@ -747,6 +744,10 @@ function! s:update_impl(pull, force, args) abort
     echohl None
   endif
 
+  let python = (has('python') || has('python3')) && !s:is_win && !has('win32unix')
+      \ && (!s:nvim || has('vim_starting'))
+  let ruby = has('ruby') && !s:nvim && (v:version >= 703 || v:version == 702 && has('patch374'))
+
   let s:update = {
     \ 'start':   reltime(),
     \ 'all':     todo,
@@ -755,7 +756,7 @@ function! s:update_impl(pull, force, args) abort
     \ 'pull':    a:pull,
     \ 'force':   a:force,
     \ 'new':     {},
-    \ 'threads': (s:py || s:ruby || s:nvim) ? min([len(todo), threads]) : 1,
+    \ 'threads': (python || ruby || s:nvim) ? min([len(todo), threads]) : 1,
     \ 'bar':     '',
     \ 'fin':     0
   \ }
@@ -768,20 +769,21 @@ function! s:update_impl(pull, force, args) abort
         \ '--depth 1' . (s:git_version_requirement(1, 7, 10) ? ' --no-single-branch' : '') : ''
 
   " Python version requirement (>= 2.7)
-  if s:py && !has('python3') && !s:ruby && !s:nvim && s:update.threads > 1
+  if python && !has('python3') && !ruby && !s:nvim && s:update.threads > 1
     redir => pyv
     silent python import platform; print(platform.python_version())
     redir END
-    let s:py = s:version_requirement(
+    let python = s:version_requirement(
           \ map(split(split(pyv)[0], '\.'), 'str2nr(v:val)'), [2, 6])
   endif
-  if (s:py || s:ruby) && !s:nvim && s:update.threads > 1
+
+  if (python || ruby) && s:update.threads > 1
     try
       let imd = &imd
       if s:mac_gui
         set noimd
       endif
-      if s:ruby
+      if ruby
         call s:update_ruby()
       else
         call s:update_python()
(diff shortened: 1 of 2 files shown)


junegunn:
Thank you. My suggestion is that we don't explicitly say anything about this change since it's a temporary workaround to a known issue. A subset of Neovim users who have python support enabled will benefit from this change, while the others will still find that vim-plug doesn't work like Vundle but they'll notice that there is an open issue. It's clearly a better situation than now anyway.

What is tricky is that there's no guarantee that `has ('vim_starting')` means that the user is expecting synchronous installation. One can just start Neovim with `+PlugUpdate` to make sure that the plugins are up-to-date. So I don't think we should disallow installation or warn the users as we can't be sure of their intentions.

plug.vim
```diff
@@ -775,6 +776,18 @@ function! s:update_impl(pull, force, args) abort
...
+  if has('nvim') && has('vim_starting')
+    if s:py
+      let s:nvim = 0
+    else
+      echohl WarningMsg
+      echomsg 'vim-plug: +PlugInstall/+PlugUpdate synchronous only with python support'
+      echomsg 'vim-plug: pip(3) install neovim'
+      echohl None
+    endif
+  endif
```
junegunn: https://github.com/junegunn/vim-plug/issues/104#issuecomment-98295463

`vim +PlugUpdate` will cause any subsequent install/update to become synchronous. I'd prefer to remove `s:` variables like I did in https://github.com/junegunn/vim-plug/issues/104#issuecomment-98295463.

starcraftman: @junegunn Right you are, will patch that once I've made the fix for the seg fault. I think it was just a matter of locking. Might have over zealously removed some.

starcraftman:
@junegunn When you say don't explicitly say anything do you mean also remove my warning? Or I'll leave it?

If you ask me, most users of +PlugInstall/Update will want synchronous behaviour because most use it as part of script/docker etc...

starcraftman:
New problem, during update with nvim ultisnips stops update?

```vim
Error detected while processing function UltiSnips#CursorMoved..provider#python#Call:                             
line   12:                                                                                                        
Invalid channel "1"  
```

plug.vim
```diff
@@ -763,25 +763,27 @@ function! s:update_impl(pull, force, args) abort
...
     redir => pyv
     silent python import platform; print(platform.python_version())
     redir END
-    let s:py = s:version_requirement(
+    let python = s:version_requirement(
           \ map(split(split(pyv)[0], '\.'), 'str2nr(v:val)'), [2, 6])
   endif
-  if (s:py || s:ruby) && !s:nvim && s:update.threads > 1
+
+  if (python || ruby) && (!s:nvim || (has('nvim') && has('starting'))) && s:update.threads > 1
```
junegunn: Let's move the second condition to line 747, the definition of `python`. It doesn't apply to `ruby`.

junegunn:
> When you say don't explicitly say anything do you mean also remove my warning? Or I'll leave it?

I meant removing warning as well, but either is fine. Your call.

> If you ask me, most users of +PlugInstall/Update will want synchronous behaviour because most use it as part of script/docker etc...

True, but we have to assume that there can be exceptional use cases.

starcraftman:
Hmmm, found new small bug. On neovim only, update gets interrupted by ultisnips:

```vim
Error detected while processing function UltiSnips#CursorMoved..provider#python#Call:                             
line   12:                                                                                                        
Invalid channel "1" 
```

starcraftman:
@junegunn Think I've gotten all the problems. Currently no warning printed to users, simply uses python installer silently. 

Limitations still existing:
* No ctrl-c/interrupt support on nvim.
* Graphical bug: neovim/python-client#103

plug.vim
```diff
@@ -747,6 +744,10 @@ function! s:update_impl(pull, force, args) abort
     echohl None
   endif
 
+  let python = (has('python') || has('python3'))
+      \ && (!s:nvim || (has('nvim') && has('vim_starting'))) && !s:is_win && !has('win32unix')
+  let ruby = has('ruby') && !has('nvim') && (v:version >= 703 || v:version == 702 && has('patch374'))
```
junegunn: How about we check `has('nvim')` only once when defining `s:nvim` and see `s:nvim` instead in other places? I think we can reorganize the expressions as follows to make them slightly less confusing

```vim
  let python = (has('python') || has('python3')) && !s:is_win && !has('win32unix')
      \ && (!s:nvim || has('vim_starting')) 
  let ruby = has('ruby') && (v:version >= 703 || v:version == 702 && has('patch374'))
      \ && !s:nvim
```

- has python & not windows & (neovim installer n/a or vim is starting)
    - We can just strip out vim_starting part when jobwait is ready
- has ruby & vim version not too old & neovim installer n/a

plug.vim
```diff
@@ -763,25 +764,27 @@ function! s:update_impl(pull, force, args) abort
   call s:prepare()
   call append(0, ['', ''])
   normal! 2G
+  redraw
```
junegunn: Is this required?

junegunn:
I commented on the code. Just some minor nitpicking. Other than those, looks great. You can squash the commits so I can merge. Thanks!

plug.vim
starcraftman: @junegunn I believe I used that line when I first started porting & the python was still crashing. I was checking the side pane code worked. The side pane created doesn't appear until a redraw is issued in neovim. Not necessary now that everything works.

plug.vim
starcraftman: @junegunn Agreed. Less messy now.

starcraftman:
@junegunn Changes made, overall a nice improvement. Hope nvim users notice.

junegunn:
@starcraftman Looks like travis build is getting delayed. I'll merge when it's done :)

junegunn:
Merged, many thanks as always!

Issue #231, opened 2015-05-19
Add 'run' to options


thebillywayne:
##### Abstract
It would be convenient to add another option for the Plug lines; namely, the ability to execute a certain command or external script subsequent to an 'on' or 'for' option. This would allow users a central location for customizing behavior based upon the selective loading of certain plugins.

##### Rationale
Many times I'd like certain commands to be executed upon the loading of certain plugins. Adding this functionality to vim-plug represents a elegant, compact solution. 

##### Example Use Case
I use Neocomplete. Neocomplete is loaded on certain filetypes. The Plug command I have is
```Plug 'Shougo/neocomplete.vim', {'for': ['python', 'scheme', 'shell', 'clojure']}```
This works very well.  
However, I have certain custom keymappings which are specific to Neocomplete. For filetypes other than those specified for Neocomplete, these  keymappings, do not apply, creating a conflict, resulting in error messages such as
* ```E117: Unknown function:Neocomplete#smart_close_popup```                                                                 
* ```E15: Invalid expression: neocomplete#smart_close_popup()."\^H"```

I realize that a workaround would be an autocommand that set the appropriate keymappings upon loading those same filetypes. This, however, leads to duplicate specifications scattered over the config files.  
An elegant solution for vimplug would be something along the lines of:
```Plug 'Shougo/neocomplete.vim', {'for': ['python', 'scheme', 'shell', 'clojure'], 'run': 'myfunction'}```  
For my case, the "function" would simply be to source a file containing my custom NeoComplete keybindings.  
I think that this enhancement would be a great feature to have in vim-plug.

**Comment by junegunn** [comment]
Hi, thanks for the suggestion. Looks like the suggested feature makes more sense for `on` option rather than for `for` option where it's already possible with `FileType` autocmd as you described above. However in this case, using `User` autocmd that is triggered only on deferred loading seems like a better and a less intrusive approach than to add an option which only benefits a few advanced users. By using autocmd, you don't have to define a function for taking its reference. And it's much simpler to implement. (A single line of `doautocmd` will do)

```vim
Plug 'Shougo/neocomplete.vim', {'for': 'xxx', 'on': 'Xxx'}
autocmd! User neocomplete.vim inoremap foo bar
```

thebillywayne:
This is a reasonable solution. I fully understand opting for a less intrusive approach. I was unaware of the `User` autocmd. I've tested it and it works very well. Thanks for the pointer.

vheon:
[...]

PR #232, opened 2015-05-19
Trigger User autocmd when plugin is loaded on demand (#231)

junegunn:


vheon:
Looks like a good thing for not lot of a hassle :+1:

junegunn:
@vheon Thanks for the review!

PR #233, opened 2015-05-19
Specify a format when `git show` displays a commit

itspriddle:
If a user has a custom `format.pretty` set in their gitconfig, `git show` will use it. This can cause display issues when viewing a commit after a PlugUpdate.

If you use `git show --pretty=medium` the diff displays properly. Medium is the default format when a custom format isn't set by gitconfig.

Example gitconfig:

```
[format]
  pretty = format:%C(red)%h%Creset%C(yellow)%d%Creset %C(magenta)%an%Creset: %s %C(green)(%ar)%Creset
```

vim-plug display before this change:

screenshot 2015-05-19 12 15 40

vim-plug display after this change:

screenshot 2015-05-19 12 16 41

### DIFF
--- a/plug.vim
+++ b/plug.vim
@@ -1934,7 +1934,7 @@ function! s:preview_commit()
   execute 'pedit' sha
   wincmd P
   setlocal filetype=git buftype=nofile nobuflisted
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks! :+1:

PR #235, opened 2015-05-23
Simple dependency.

starcraftman:
Probably seems simple to you @junegunn but some people as I saw in an r/vim thread a while ago thought this neat. Simple way to manage dependencies of plugins that Plug can fetch.

By having it on one line, easily delete or comment plugin + deps.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,9 @@ call plug#begin('~/.vim/plugged')
 Plug 'junegunn/seoul256.vim'
 Plug 'junegunn/vim-easy-align'
 
+" Group dependencies, vim-snippets depends on ultisnips
+Plug 'SirVer/ultisnips' | Plug 'honza/vim-snippets'
+
 " On-demand loading
 Plug 'scrooloose/nerdtree', { 'on':  'NERDTreeToggle' }
 Plug 'tpope/vim-fireplace', { 'for': 'clojure' }
--- a/plug.vim
+++ b/plug.vim
@@ -14,6 +14,9 @@
 "   Plug 'junegunn/seoul256.vim'
 "   Plug 'junegunn/vim-easy-align'
 "
+"   " Group dependencies, vim-snippets depends on ultisnips
+"   Plug 'SirVer/ultisnips' | Plug 'honza/vim-snippets'
+"
 "   " On-demand loading
 "   Plug 'scrooloose/nerdtree', { 'on':  'NERDTreeToggle' }
 "   Plug 'tpope/vim-fireplace', { 'for': 'clojure' }


vheon:
I don't know if it would cause confusion. I mean, I don't know if it is clear enough that it is the same as doing 

```vim
Plug 'Sirver/Ultisnips'
Plug 'honza/vim-snippets'
```

and that vim-plug does nothing special about it, no check, no special ordering etc etc.

starcraftman:
@vheon True enough, it's mainly just an organizational trick for the vimrc. We can just as easily not add this line. I just thought some people might like.

vheon:
I'm not saying that it isn't cool, I'm saying that it feels wrong calling it just "simple plugin dependency". Maybe note that it is possible that two or more plug command can be inlined but that it is just a stylistic matter.

junegunn:
@starcraftman Hey, thanks. I kind of like it because the `|` reminds me of the expression for conditional probability. I think @vheon has a point. How about labeling the line slightly differently, like "Inlining dependent plugins" or something similar.

starcraftman:
@junegunn I agree, that does summarize it better.  I made it: `Inline dependent plugins by ORing` to be absolutely clear.

junegunn:
@starcraftman Thanks but I'm not sure if "OR" is the right term to use here. The bar character (`:help bar`), though used for bitwise-or in some languages, has nothing to do with "OR" in this context. I also feel that the use of imperative present tense may give the wrong message that they "have to be inlined". Sorry for being picky :) What do you think?

starcraftman:
@junegunn Re verb vs adjective, yes, inline is technically an adjective but it has entered verb usage in my opinion, just like Googling. Also, I see what your saying about OR, guess that is a bit misleading. 

I like your suggestion, I changed it slightly to `Inline declaration of plugin dependencies` because I think people are more familiar with that phrasing over the word dependent.

vheon:
>  I changed it slightly to Inline declaration of plugin dependencies because I think people are more familiar with that phrasing over the word dependent

I don't know if it will be clear that the dependency direction is the opposite of the one might think. Instead using what @junegunn suggested it is more clear that the one that we are inlining are plug dependant of the plug mentioned before. So for example in:

```viml
Plug 'junegunn/vim-pseudocl' | Plug 'junegunn/vim-fnr'
```

`vim-fnr` is dependant on `vim-pseudocl` and not the other way around. It is rare that order matter in plugin inclusion, but in case the dependant **has** to be loaded after the dependency, it has to be clear how to use this syntax. Some people could not know it and think that is a special thing of vim-plug.

starcraftman:
@vheon I see what your saying, I guess I was over thinking it. 

As one last alternative, do you find it clearer if I just spell out the example?

```
" Group dependencies: vim-snippets DEPENDS on ultisnips
Plug 'SirVer/ultisnips' | Plug 'honza/vim-snippets'
```

Side note: What an exercise is the English language :).

vheon:
Yes, I prefer this one :+1: 

> Side note: What an exercise is the English language :).

Yeah, I'm not a native English speaker, so it's even harder for me :stuck_out_tongue_closed_eyes:

starcraftman:
@junegunn Unless you've any objection, I think the wording is now quite clear.

junegunn:
@starcraftman @vheon Thanks, guys!

PR #236, opened 2015-05-27
Bugfix of on-demand loading

hori-ryota:
Using `Plug 'scrooloose/nerdtree', { 'on':  'NERDTreeToggle' }` and reload `.vimrc`, nerdtree isn't contained `s:loaded` and cannot be rehooked.

So, I change this script for `s:loaded` containing on-demand loaded plugins when `.vimrc` is reloaded.

### DIFF
--- a/plug.vim
+++ b/plug.vim
@@ -440,7 +440,7 @@ function! s:add(repo, ...)
       call add(g:plugs_order, name)
     endif
     let g:plugs[name] = spec
-    let s:loaded[name] = 0
+    let s:loaded[name] = get(s:loaded, name, 0)
   catch
     return s:err(v:exception)
   endtry


junegunn:
Thanks. Can you elaborate what "cannot be rehooked" means? Maybe an example? (By the way, I think some test will fail as they assume that the loaded flag is reset on another `Plug` command)

hori-ryota:
Using `Plug 'scrooloose/nerdtree', { 'on':  'NERDTreeToggle' }`,
execute `NERDTreeToggle` hooks loading NERDTree,
reloading .vimrc remove nerdtree from `s:loaded`,
and re-execute `NERDTreeToggle` doesn't hook loading NERDTree because `exists(:NERDTreeTogge)` returns `2` ([vim-plug/plug.vim at master · junegunn/vim-plug](https://github.com/junegunn/vim-plug/blob/master/plug.vim#L175)).

Even if this existing check removed, `NERDTreeToggle` will become undefined because `g:loaded_nerd_tree` is exists.

So, I think that we need redefine NEADTreeToggle in plug.vim or keep loaded flags.

hori-ryota:
Using `Plug 'scrooloose/nerdtree', { 'on': 'NERDTreeToggle' }`,
execute NERDTreeToggle hooks loading NERDTree,
reloading `.vimrc` remove nerdtree from `s:loaded`,
and re-execute `NERDTreeToggle` doesn't hook loading NERDTree because `plug.vim` doesn't override `NERDTreeToggle` by reason of `exists(:NERDTreeTogge)` returning `2` ([vim-plug/plug.vim at master · junegunn/vim-plug](https://github.com/junegunn/vim-plug/blob/master/plug.vim#L175)).

Even if this existing check removed, `NERDTreeToggle` will become undefined because `g:loaded_nerd_tree` is exists.

So, I think that we need redefine original `NEADTreeToggle` in `plug.vim` or keep loaded flags.

junegunn:
I have the exact line in [my vimrc](https://github.com/junegunn/dotfiles/blob/master/vimrc#L65) and I have no problem loading nerdtree even when .vimrc is reloaded. Am I missing something?

1. Start vim
1. `:command NERD`

    ```vim
        Name        Args Range Complete  Definition
    !   NERDTreeToggle *    .               call s:lod_cmd('NERDTreeToggle', "<bang>", <line1>, <line2>, <q-args>, ['nerdtree'])
    ```
1. `:source ~/.vimrc`
1. `:command NERD`

    ```vim
        Name        Args Range Complete  Definition
    !   NERDTreeToggle *    .               call s:lod_cmd('NERDTreeToggle', "<bang>", <line1>, <line2>, <q-args>, ['nerdtree'])
    ```
1. `:NERDTreeToggle` correctly opens nerdtree
1. `:command NERD`

    ```vim
        Name        Args Range Complete  Definition
        NERDTree    ?          dir       :call g:NERDTreeCreator.CreatePrimary('<args>')
        NERDTreeCWD 0                    call NERDTreeCWD()
        NERDTreeClose 0                    :call g:NERDTree.Close()
        NERDTreeFind 0                    call s:findAndRevealPath()
        NERDTreeFocus 0                    call NERDTreeFocus()
        NERDTreeFromBookmark 1          customlist call g:NERDTreeCreator.CreatePrimary('<args>')
        NERDTreeMirror 0                    call g:NERDTreeCreator.CreateMirror()
        NERDTreeToggle ?          dir       :call g:NERDTreeCreator.TogglePrimary('<args>')
    ```

hori-ryota:
Please add `:NERDTreeToggle` before 3 (`source ~/.vimrc`) and try `:PlugUpdate` after 5 (`:NERDTreeToggle`) then nerdtree shows not loaded.

hori-ryota:
`NERDTreeToggle` command is enable after 5 (`:NERDTreeToggle`) but nerdtree isn't contained `s:loaded` and nerdtree directory isn't contained `runtimepath`.

So, to give an example of one impact point, NERDTree syntax is broken after `source ~/.vimrc`.

junegunn:
Ah now I get it, thanks. Could you change your code to `let s:loaded[name] = get(s:loaded, name, 0)`? It's more idiomatic in vim-plug's source code. And before merging your code to master, I'd like to fix the failing tests and add a test case for this particular scenario, I'll look into it later in the day.

junegunn:
Thanks, but please squash the commits into one.

junegunn:
Merged to master, thanks!

PR #242, opened 2015-06-10
Remove GLog.

starcraftman:
Does what it says. The py path seems stable enough to remove it now. Just dead code taking up lines.


junegunn:
Thanks :+1:

PR #252, opened 2015-06-25
Remap window close to Q

elken:
Remap `q` to `Q`, as `q` requires and extra key press, and it matches the capitalized scheme of the other keys. Feel free to map to something else, `Q` was just a suggestion as its unlikely someone would use `ex` mode in a plugin buffer. :smiley:

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,6 +99,7 @@ Reload .vimrc and `:PlugInstall` to install plugins.
 | `g:plug_shallow`    | 1                                 | Use shallow clone                                      |
 | `g:plug_window`     | `vertical topleft new`            | Command to open plug window                            |
 | `g:plug_url_format` | `https://git::@github.com/%s.git` | `printf` format to build repo URL                      |
+| `g:plug_quit`       | `q`                               | Key to close the window                                |
 
 ### Keybindings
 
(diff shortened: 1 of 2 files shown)


starcraftman:
> Remap q to Q, as q requires and extra key press
I'm not sure I understand this rationale. Q requires holding down shift + q so that's an extra key press. 

I get the other bit about all other maps having caps. Personally I like lowercase.

elken:
I suppose. The rest of the motions are all upper case, so it makes sense to have a consistent scheme.

vheon:
The thing is that `q` is the default for quitting paging like `less` so people have already the mental model for that.

elken:
Fine, I changed it to be optional.

junegunn:
Thanks but we can't change something people have been using for more than a year without good reason. You do have a point, but I would still pick the lowercase `q`, because it's the de-facto standard "close key" for vim plugins. It's the community convention. For those who really want to override the settings of plug window, there's `FileType` autocmd, so the new option is not necessary.

```vim
autocmd FileType vim-plug nmap Q q
```

elken:
Wow, I can't believe I forgot the au....

Sorry Junegunn.

elken:
Wow, I can't believe I forgot the au....

Sorry Junegunn.

junegunn:
No problem.

PR #255, opened 2015-07-07
PlugDiff mappings

sodapopcan:
Hoping you will be interested in this (obviously) and maybe try it out a bit (I've only been using it for a day).

This introduces some new mappings for `:PlugDiff`:

`J` and `K` can be used from the diff buffer to scroll the commit preview window without having to switch to it.

`<C-N>` and `<C-P>` can be used to jump straight to commit lines (sorta like jumping to files in fugitive's commit window).

I've also made the commit window `unmodifiable` and mapped `q` to `:quit` (while inside it).

This PR introduces a `s:buffocus()` utility function.  This is used to ensure that the `J` and `K` mappings will only ever scroll the commit preview window.  If you would like me to ditch it in favour of `wincmd p`--or just change its location in the file--I will oblige.

I've also documented the `<CR>` and `o` mappings for `:PlugDiff`.

If you're willing to accept this PR, I'm happy to make any other changes or edits but otherwise, thanks for this amazing plugin manager.

### DIFF
--- a/plug.vim
+++ b/plug.vim
@@ -1904,6 +1904,8 @@ function! s:preview_commit()
   setlocal filetype=git buftype=nofile nobuflisted
   execute 'silent read !cd' s:shellesc(g:plugs[name].dir) '&& git show --pretty=medium' sha
   normal! gg"_dd
+  setlocal nomodifiable
+  nnoremap <silent> <buffer> q :q<cr>
   wincmd p
 endfunction
 


junegunn:
Thanks, looks useful. I'll review the code later in the day. Can you write test cases for these?

junegunn:
By the way, the same idea can be implemented without touching the core. This approach can be considered to be more flexible as it allows users to choose whatever keys they want. (e.g. `CTRL-J/K` instead of `J/K`) What do you think?

```vim
function! s:scroll_preview(down)
  silent! wincmd P
  if &previewwindow
    execute 'normal!' a:down ? "\<c-e>" : "\<c-y>"
    wincmd p
  endif
endfunction

function! s:setup_extra_keys()
  nnoremap <silent> <buffer> J :call <sid>scroll_preview(1)<cr>
  nnoremap <silent> <buffer> K :call <sid>scroll_preview(0)<cr>
  nnoremap <silent> <buffer> <c-n> :call search('^  \zs[0-9a-f]')<cr>
  nnoremap <silent> <buffer> <c-p> :call search('^  \zs[0-9a-f]', 'b')<cr>
endfunction

autocmd! FileType vim-plug call s:setup_extra_keys()
```

sodapopcan:
That's a lot more elegant than mine :+1:  I know I'm doing a bit much with the `s:buffocus()` to cover an edge case... I got overboard sometimes!

I chose `J/K` since it's very common for people to use `CTRL-H/J/K/L` for window navigation and I always appreciate it when the plugin makes the choices for mappings in unmodifiable buffers (as vim-plug does!)  This is the first time I've ever found myself wanting to scroll and alternate window, so I wanted some new mappings for that idea.

So you're just saying to use this in my vimrc?  If you don't want to include it, is it cool to at least make the preview window `unmodifiable`?  I'm always on a 13" laptop so I find myself jumping into it and pressing `q` to close it in order to see more of the update list.  I might be alone in this.

The other thing I was thinking `<C-N>/<C-P>` could do is jump to the first commit of the next plugin.  Not sure how you feel about that.

In any event, thanks for the attention to this!  I know I'm talking a lot about a small feature, but `:PlugDiff` is one of my favourite aspects of this plugin.

And yes, I can write tests for anything you are willing to accept!

junegunn:
+1 to `unmodifiable` and `q`. I'm a bit split on the other ones though. In principle, we know that the "right" way to have them is to provide `<Plug>` mappings for each one of them so that the users can customize if they do not like the default. But at the same time, I'm not a big fan of plugins with too many configuration knobs. And since I'm not certain yet which set of keys or behaviors should be the default, I'm leaning towards just putting the above code on the wiki page, so that the users can have it on their vimrcs however they like.

sodapopcan:
I'm fine with all that.  Thank you, sir!

junegunn:
Thanks, could you to update the PR to have only `q` and `unmodifiable`? The change becomes trivial then but I'd like to give you some credit for it.

sodapopcan:
Done.  Thanks a lot!

junegunn:
Merged, thanks! I'll create a wiki page and let you know.

junegunn:
So here's the page: https://github.com/junegunn/vim-plug/wiki/extra

You can freely update or add examples. Thanks.

sodapopcan:
Awesome, thanks for the credit :)

`gx` is also a great idea!

junegunn:
FYI, I just realized that making the buffer `nomodifiable` caused problem when `o` or `enter` is pressed again for the commit. Pushed the fix.

sodapopcan:
Oh crap, I never tried that.  I only tried opening different commits.  My
bad.  Thank you!

junegunn:
No problem!

PR #270, opened 2015-08-22
Add local plugin usage to the README

xanderdunn:
This was added to the README in
466d1839b37c174a66b329fadce2512b0a2ffa5e, but for some reason the
functionality is still there but it’s no longer documented in the
README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,9 @@ Plug 'tpope/vim-fireplace', { 'for': 'clojure' }
 " Using git URL
 Plug 'https://github.com/junegunn/vim-github-dashboard.git'
 
+" Local plugin
+Plug '/my/local/vim/plugin'
+
 " Plugin options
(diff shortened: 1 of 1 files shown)


junegunn:
Hi, thanks for the PR but I think you missed it somehow.
See https://github.com/junegunn/vim-plug/blame/master/README.md#L68-L69
The current example with `~` I believe is a better example since it makes it possible to use the same .vimrc over different platforms and machines (e.g. `/Users/jg/xyz`, `/home/jg/xyz`, `/root/xyz`)

xanderdunn:
Aha.  I see.  I was looking for the word "local".  I think it would be easier to find if that word were in there somewhere.  It probably comes to mind more often than "unmanaged."

junegunn:
Possibly. For some reason, I thought "unmanaged" was technically more precise term for the case so I changed it here: https://github.com/junegunn/vim-plug/commit/7e55690f1940b5db21675bae91bd6f7f42df91b2

PR #273, opened 2015-08-26
Fix #272 PlugUpdate and delayed loading

### LINKED ISSUE #272: PlugUpdate and delayed loading not always playing nicely together
While exploring an issue with the [jedi-vim](https://github.com/davidhalter/jedi-vim/issues/459) plugin I encountered a possible bug:

In my [.vimrc](https://github.com/montefra/.vim/blob/vim-plug/.vimrc) I add jedi as:

    Plug 'davidhalter/jedi-vim', { 'for':  'python' }

If I open a vim session and do:

1. ``PlugUdate``
2. open a python file

the popup with the function/class/... signature looks very messy (see the screenshots in the above link, all the ``=`=jedi=0, =`=    `` should not be there). But it works without issues if I open it in a session where PlugUpdate has not been run.

If I don't delay the loading:

    Plug 'davidhalter/jedi-vim'

and repeat the above exercise, then the pop up looks fine.

My guess is that the reload at the end of ``PlugUpdate`` might not play very well with not jet loaded plugins.

starcraftman:
* Force syntax files sourcing on lod_ft.

Only negative to this fix is sometimes double sourcing syntax files, not a big deal it seems.

### DIFF
--- a/plug.vim
+++ b/plug.vim
@@ -399,7 +399,7 @@ function! s:lod(names, types)
 endfunction
 
 function! s:lod_ft(pat, names)
-  call s:lod(a:names, ['plugin', 'after/plugin'])
+  call s:lod(a:names, ['plugin', 'after/plugin', 'syntax', 'after/syntax'])
   execute 'autocmd! PlugLOD FileType' a:pat
   if exists('#filetypeplugin#FileType')
     doautocmd filetypeplugin FileType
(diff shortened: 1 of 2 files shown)


junegunn:
Thanks!

starcraftman:
@junegunn No problem. :)

PR #283, opened 2015-09-11
Refactor Python Command & .travis.yml

starcraftman:
A few highlights:
* ~~Change from time.sleep to using a thread that blocks & join it. Should reduce cpu polling problem and make it faster.~~
* A lot of style improvements in the Command class and surrounding code.
* Change travis to use env instead of many ruby versions. Makes it easier to see what build broke as env can be 'nox', 'python', 'python3', or 'ruby'. Also, I think the script reads easier.

Need to spend more time investigating the time.sleep alternative. For now, I'll just increase the sleep to reduce load.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,35 +1,49 @@
 language: ruby
 rvm:
   - 1.8.7
-  - 1.9.2 # Test with vim-nox package on ubuntu
-  - 1.9.3 # Test against python installer
   - 2.0.0
-  - 2.1.0 # Test against python3 installer
-
-before_script: |
-  sudo apt-get update -y
-  if [ $(ruby -e 'puts RUBY_VERSION') = 1.9.2 ]; then
-    sudo apt-get install -y vim-nox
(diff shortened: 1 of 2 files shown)


starcraftman:
@junegunn Any objections re travis? Otherwise, don't think there's anything you'd be bothered by.

starcraftman:
Hmmm, seems to have an impact on KeyboardInterrupt support, will have to look into that. Otherwise seems no issue.

junegunn:
Looks good to me. Love the new travis.yml. I can read and write basic Python, but I don't know well about "good style", so I'll leave that up to you. Just please make sure that there's no regression in the cases that are not (or cannot effectively) covered by the Vader test scenario, user interaction, timeout, etc. You can merge when you think it's ready.

starcraftman:
@junegunn Ok then, seems to have a small bug regarding thread shutdown that occurs once in a while. Just need to figure it out via logs then it should be good to go.

starcraftman:
All done and passing my tests now. Pushing that merge button.

PR #285, opened 2015-09-18
Join a thread instead of sleeping

starcraftman:
* Instead of having a fixed sleep duration, start a thread
that monitors the subprocess.
* Started thread is joined, if process finishes thread returns
before timeout.
* Reduces time taken & CPU overhead.

Still want to do some testing with this, but I am pretty certain I solved the earlier bug with the thread_not_started loop. Just want to test it myself for a few days before merging, including error cases. On my two comps, reduces time for PlugUpdate by a 1-2 secs.


starcraftman:
Done a few days of testing, I see no problem. Merging.

PR #286, opened 2015-09-18
Update YouCompleteMe install script

zackhsi:
The install script has been changed from `install.sh` to `install.py`.

https://github.com/Valloric/YouCompleteMe/commit/1b40d683be7fdc38bf04c2e889036314551f0035


starcraftman:
@zackhsi Thanks for the notification. Guess we should provide right examples. The spacing around the brackets seems quite jarring in the new script, odd choice.

PR #293, opened 2015-10-10
add `.nvim` note for neovim users

aeosynth:


junegunn:
Thanks, but I think it's stating the obvious. It's a little hard for me to imagine someone adventurous enough to choose neovim over vim being unaware of such a difference. And they should really pay attention to `:help vim-differences` or `:help nvim-from-vim`.

PR #309, opened 2015-10-30
Add windows install instructions

timjk:


README.md
```diff
@@ -23,14 +23,23 @@ A minimalist Vim plugin manager.
 [startup-time]: http://junegunn.kr/images/vim-startup-time.png
 
 ### Installation
-
 [Download plug.vim](https://raw.githubusercontent.com/junegunn/vim-plug/master/plug.vim)
+###### Unix
 and put it in ~/.vim/autoload
```
junegunn: Thanks, but I would rather not split the sentence since doing so will give the impression that manual downloading of the link is necessary. How about changing it to "Download plug.vim and put it in ~/.vim/autoload (or ~/vimfiles/autoload on Windows)?

DustinVenegas:
Vim on windows should pick up a "~/.vim" directory.

junegunn:
@DustinVenegas Can you elaborate? I no longer use windows, but the last time I checked it was `vimfiles` directory.

README.md
```diff
@@ -23,14 +23,23 @@ A minimalist Vim plugin manager.
...
 
 ```sh
 curl -fLo ~/.vim/autoload/plug.vim --create-dirs \
     https://raw.githubusercontent.com/junegunn/vim-plug/master/plug.vim
 ```
+###### Windows
+and put it in ~/vimfiles/autoload
+
+```sh
```
starcraftman: Shouldn't this be batch?

starcraftman:
@junegunn @DustinVenegas Technically speaking, there is nothing preventing you from making a dot directory named `.vim` and using that. It will look a bit odd, as on Windows it isn't hidden or otherwise special.

IMO though, on Windows I have always used `_vimrc` (due to gui preventing leading dot files creation) and `vimfiles` as seems convention. When I see `.vim` on Windows, I usually assume that it is configuration for a pseudo posix shell like cygwin or babun. I don't think my understanding is unique.

So yes, if we put Windows specific instructions, I'd stick to the `vimfiles` convention. People using pseudo posix will just use the unix instructions.

README.md
starcraftman: I might suggest the following a bit clearer:
"Download plug.vim and put it in `~/.vim/autoload` (UNIX Shells) or `~/vimfiles/autoload` (Windows Shell).

junegunn: I think we can omit `Shells` and `Shell`.

starcraftman: Ok, then:
"Download plug.vim and put it in `~/.vim/autoload` (UNIX) or `~/vimfiles/autoload` (Windows).

junegunn:
Thanks, I'd like to make a few minor requests on the updated revision:
- To make it more consistent with the other parts of the doc, please add
    - line break before `and put it in`
    - empty lines around `#####`
- And squash the commits into one

starcraftman:
@timjk @junegunn Additional note, the command provided to download is available only for Powershell 3.0+. That means only Windows 8.0 or later, and even then, excludes standard cmd prompt used by many older guys. Should be indicated.

1. It should be indicated, so please change the section head from `Windows` to  `Windows Powershell 3.0+`
1. I did a bit of experimenting, you can support Powershell 2.0+ (available on XP via install & 7 I believe default) with the following snippet. I'd suggest that's at least more powershells support.

```powershell
md ~\vimfiles\autoload

(New-Object Net.WebClient).DownloadFile('https://raw.githubusercontent.com/junegunn/vim-plug/m
aster/plug.vim', 'vimfiles\autoload\plug.vim')
``

Batch unfortunately left by the way side.

**Comment by timjk** [comment]
I'm happy to update the readme to reflect the powershell 2.0+ version it is visually less elegant but most users will be copy pasting so it shouldn't matter.

@junegunn Thoughts? I will also be adding in your requests.

**Comment by starcraftman** [comment]
@timjk I think your right, it isn't super likely Windows people would confuse the two shells. Still use the older command for more support, just ignore my comment re labeling. I'll just edit the README myself if there's a torrent of batch users saying "Why no work?". :)

**Comment by junegunn** [comment]
- `Windows` instead of `Windows Powershell`. I truly believe that most Windows users don't even know what Powershell is :)
- More compatible snippet (the one @starcraftman suggested, right?)
- Minor nit picks I [commented above](https://github.com/junegunn/vim-plug/pull/309#issuecomment-153281709)

**Comment by DustinVenegas** [comment]
The target Windows audience using vim-plug should be familiar with Powershell. Chocolatey has a very similar installer type. I would target Powershell 2 though. Anyone that can't figure out the shell can always install vim-plug manually.

**Comment by mooreniemi** [comment]
FWIW, I found the powershell instructions to be exactly what I wanted and very easy. Thank you @DustinVenegas.

**Comment by starcraftman** [comment]
@mooreniemi Good to know. I do want to merge this, but @timjk seems a bit occupied and hasn't made the requested changes to his branch. I'll give him a few more days, if not I'll just merge it manually and amend his commit.

**Comment by junegunn** [comment]
@starcraftman Agreed. I'll be mostly unavailable for the rest of November due to vacation. I'll leave this up to you.

**Comment by starcraftman** [comment]
@junegunn No problem, I'll deal with it later today if time permits. I think I've waited long enough.

Have a nice vacation!

**Comment by starcraftman** [comment]
@timjk Thanks for the initial PR, I merged it to master today with fixes discussed.

**Comment by mooreniemi** [comment]
Hmm, I might've been Doing It Wrong but when I tried the merged instructions I got an error.

```Windows PowerShell
Copyright (C) 2014 Microsoft Corporation. All rights reserved.

PS C:\Windows\system32> md ~\vimfiles\autoload


    Directory: C:\Users\quartz\vimfiles


Mode                LastWriteTime     Length Name
----                -------------     ------ ----
d----        11/29/2015   2:52 PM            autoload


PS C:\Windows\system32> $uri = 'https://raw.githubusercontent.com/junegunn/vim-p
PS C:\Windows\system32> (New-Object Net.WebClient).DownloadFile($uri, 'vimfiles\
Exception calling "DownloadFile" with "2" argument(s): "An exception occurred du
At line:1 char:1
+ (New-Object Net.WebClient).DownloadFile($uri, 'vimfiles\autoload\plug.vim')
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [], MethodInvocationException
    + FullyQualifiedErrorId : WebException```

The original instructions below worked for me though:

```PS C:\Windows\system32> Invoke-WebReq

cmdlet Invoke-WebRequest at command p
Supply values for the following param
Uri: https://raw.githubusercontent.co
PS C:\Windows\system32>
```

starcraftman:
@mooreniemi If I'm reading your shell right, you are operating with current working directory set to `C:\Windows\system32`, then the download is trying to put plug.vim into a folder relative that cwd, that is: `C:\Windows\system32\vimfiles\autoload\plug.vim`, which doesn't exist I assume. The oversight here is I should have resolved the path on the destination, I'll have to update that command momentarily.

A few questions:
1. I assume ~ expands to your actual user home, not system32? Try `Resolve-Path ~/vimfiles/autoload/plug.vim`.
2. Is your powershell set to start your working directory in system32 always? Or were you just there when you decided to use the command?
3. Can you confirm for me that this command works: `(New-Object System.Net.WebClient).DownloadFile($uri, (Resolve-Path ~\vimfiles\autoload\plug.vim))`

Sidenote:
Powershell's error message should say somewhere "Target folder doesn't exist", instead they tell you a bunch of other stuff that is completely useless. Award winning design.

mooreniemi:
@starcraftman Yeah, that was totally it -- my bad. I just misinterpreted the Powershell output. Thanks for the help!

PR #321, opened 2015-11-23
Add install directions for using neovim

xanderdunn:
Neovim's autoload directory path is now considerably different from vim's.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,12 @@ curl -fLo ~/.vim/autoload/plug.vim --create-dirs \
     https://raw.githubusercontent.com/junegunn/vim-plug/master/plug.vim
 ```
 
+If you're using neovim:
+```sh
+curl -fLo ~/.config/nvim/autoload/plug.vim --create-dirs \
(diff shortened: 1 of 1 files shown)


starcraftman:
@xanderdunn Uh, that is annoying. Not your PR but neovim's config change. Now means we will have 3 ways in install. More to the point now cluttering `.config` and forcing me to type more directories to get to stuff. Nuissance seems to about sum it up.

README.md
```diff
@@ -34,6 +34,12 @@ curl -fLo ~/.vim/autoload/plug.vim --create-dirs \
     https://raw.githubusercontent.com/junegunn/vim-plug/master/plug.vim
 ```
 
+If you're using neovim:
```
starcraftman: Please make it like the surrounding sections, so this line should be like below. Followed by a space.

`###### Neovim`

xanderdunn:
@starcraftman Thanks for looking at this.  I pushed an additional commit.  That should do it.

starcraftman:
@xanderdunn Please just squash the second commit into the first. Otherwise fine.

starcraftman:
Did it manually.

PR #324, opened 2015-11-24
Fix #318: Freezes on PlugInstall

### LINKED ISSUE #318: Freezes on PlugInstall
vim freezes when I do `PlugInstall`. Strangely, it does not freeze when I run it from a fresh install.

I set `g:plug_threads = 1` and that has resolved it. Is it because I have vim compiled without ruby? (It is compiled with python and python3). I was under the impression from the documentation that ruby isn't required and that python multithreading would be used if it wasn't available?

This is on Ubuntu 14.04 in tmux.

starcraftman:
Fixes the issue, need to be careful with threads during try blocks.

### DIFF
--- a/plug.vim
+++ b/plug.vim
@@ -1061,7 +1061,7 @@ endwhile
 endfunction
 
 function! s:update_python()
-let py_exe = has('python3') ? 'python3' : 'python'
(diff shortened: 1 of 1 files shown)


junegunn:
:+1:

PR #329, opened 2015-11-26
Add install directions for using neovim

starcraftman:
Neovim's autoload directory path is now considerably different from vim's.

Got impatient, squashed patch myself.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,13 @@ curl -fLo ~/.vim/autoload/plug.vim --create-dirs \
     https://raw.githubusercontent.com/junegunn/vim-plug/master/plug.vim
 ```
 
+###### Neovim
+
+```sh
+curl -fLo ~/.config/nvim/autoload/plug.vim --create-dirs \
+    https://raw.githubusercontent.com/junegunn/vim-plug/master/plug.vim
+```
+
 ###### Windows
 
 ```powershell


junegunn:
Thanks. Should we update line 27-28 as well accordingly? Or how about replacing `##### Neovim` with a short note that neovim uses different path? The thing is `Unix`, `Neovim`, and `Windows` don't seem to be in the same category.

starcraftman:
@junegunn Good point, its a bit of a duplication between that section and below. Should 27-28 maybe it should just read: 

[Download plug.vim](https://raw.githubusercontent.com/junegunn/vim-plug/master/plug.vim) and put it in your autoload directory.

Alternatively we could convert it to a list:

[Download plug.vim](https://raw.githubusercontent.com/junegunn/vim-plug/master/plug.vim) and put it in the appropriate autoload directory.
- UNIX: `~/.vim/autoload`
- Neovim: `~/.config/nvim/autoload`
- Windows: `~/vimfiles/autoload`.

Regarding the UNIX/Neovim clash, don't know how to resolve that. While it is out of place, calling first section Vim with two subsections UNIX/WIndows seemed wrong (i.e. useless nesting). I'm open to suggestions. The problem here is neovim is diverging from standard, wonder where the Windows config will end up later....

The original PR used a short note, but the small text seems to clash surrounded by two larger code blocks.

junegunn:
Neovim will eventually support Windows. I wonder what the path should be then. Four pretty much identical subsections? :/

Anyway, for now, let's just change the line 27-28 to "Download plug.vim and put it in your autoload directory." and be done with it.

starcraftman:
@junegunn I just checked their latest [Windows build instructions](https://github.com/neovim/neovim/wiki/Installing-Neovim#vimrc-file-in-windows) they've decided on:
`%userprofile%\AppData\Local\nvim\init.vim`. Uhhh.... personal dislike of AppData area on Windows, seems about as organized as most peoples $HOME. I'm not sure I see any real point to changing the standard dirs, except to confuse users.

PR #334, opened 2015-11-28
Migrate travis to use new docker system.

starcraftman:
* test/run now detects best vim to use.
* travis now installs built vim into $DEPS (or $HOME/deps).

Advantages:
- Faster startup/execution.
- Travis a bit more explicit.
- Don't know if legacy support will ever be dropped, preventative.

Cons: 
- Sudo still unsupported, but don't think we will ever need.

I'll just leave this for your feedback @junegunn, if you are alright with changes its fine to go.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,49 +1,51 @@
 language: ruby
-rvm:
-  - 1.8.7
-  - 2.0.0
+sudo: false
 env:
-  - ENV=nox
-  - ENV=python
-  - ENV=python3
-  - ENV=ruby
+  global:
+    - DEPS=$HOME/deps
+    - PATH=$DEPS/bin:$PATH
 matrix:
-  exclude:
-    - rvm: 2.0.0
   include:
-    - rvm: 2.0.0
-      env: ENV=ruby
+    - env: ENV=nox
+      rvm: 1.8.7
+      addons: { apt: { packages: [vim-nox] } }
+    - env: ENV=python
+      rvm: 1.8.7
+      addons: { apt: { packages: [python2.7-dev] } }
+    - env: ENV=python3
+      rvm: 1.8.7
+      addons: { apt: { packages: [python3-dev] } }
+    - env: ENV=ruby
+      rvm: 1.8.7
+    - env: ENV=ruby
+      rvm: 2.0.0
 install: |
   git config --global user.email "you@example.com"
   git config --global user.name "Your Name"
-  sudo apt-get update -y
 
   if [ "$ENV" == "nox" ]; then
-    sudo apt-get install -y vim-nox
-    sudo ln -s /usr/bin/vim /usr/local/bin/vim
+    mkdir -p ${DEPS}/bin
+    ln -s /usr/bin/vim.nox ${DEPS}/bin/vim
     return
   fi
 
-  C_OPTS="--with-features=huge --disable-gui "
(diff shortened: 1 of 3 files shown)


starcraftman:
On a related note, I have noticed intermittent build failures happening at random. After your doc change yesterday, I had to do several restarts to get a green build. I guess it is just travis/github being flakey but should keep in mind.

.travis.yml
```diff
@@ -1,37 +1,39 @@
...
-  C_OPTS="--with-features=huge --disable-gui "
+  C_OPTS="--prefix=$DEPS --with-features=huge --disable-gui "
   case "$ENV" in
     python)
-      PACKS=python2.7-dev
       C_OPtS+=--enable-pythoninterp
       ;;
     python3)
-      PACKS=python3-dev
       C_OPtS+=--enable-python3interp
```
junegunn: Hmm, I missed this the last time, but the lowercase `t` here is not intentional, right?

test/run
```diff
@@ -85,13 +85,26 @@ DOC
   cd "$BASE"
 }
 
+select_vim() {
+  local vim=/usr/bin/vim
+  if [ "${DEPS}x" != "x" ] && [ -e "${DEPS}/bin/vim" ]; then
```
junegunn: How about using `[ -n "${DEPS}" ]`?

.travis.yml
starcraftman: Ooops.

test/run
starcraftman: Sure, just my own personal bash preference.

junegunn:
One more thing: [their article](https://blog.travis-ci.com/2014-12-17-faster-builds-with-container-based-infrastructure) says that `sudo: false` is required to use the container-based stack, but I don't see it here Something's changed, or you just forgot to put it there?

starcraftman:
@junegunn Ah that's true, hehe. Also geez, github really having problems. One of the last travis jobs got a 443 on the clone :/.

starcraftman:
Funny how we both missed that glaring typo, all fixed up now. Any other comments?

plug.vim
```diff
@@ -1283,7 +1283,7 @@ class Plugin(object):
         with self.lock:
           thread_vim_command("let s:update.new['{0}'] = 1".format(self.name))
     except PlugError as exc:
-      self.write(Action.ERROR, self.name, str(exc))
+      self.write(Action.ERROR, self.name, [str(exc)])
```
junegunn: :question:

starcraftman: Uncaught bug in the py installer due to aforementioned glaring typo.

junegunn: Okay.

junegunn:
Looks like the overhead of the legacy mode of execution was not a dominant factor in build time and thus the change doesn't seem to make the build *noticeably faster*. But still I prefer the style of the new travis.yml, so let's go for it. Thanks!

starcraftman:
@junegunn One advantage of new infrastructure I forgot to mention is optional caching. Though I didn't enable it as I'm not sure what kind of cache expiration policy we want. By default, cache lasts until you manually expunge it via dashboard. If we don't care about most recent vim build, we could just invalidate it ourselves when we feel like it.

junegunn:
Ah, okay, thanks for the explanation. Sounds nice, but rapid build is obviously not something we desperately need at the moment. :)

PR #337, opened 2015-11-29
Ensure CWD isn't a factor for powershell command.

starcraftman:
Just expands the path to ensure it downloads to directory we specified to md. Just had a question where user wasn't in right CWD.


junegunn:
Thanks!

PR #342, opened 2015-12-03
Support multi-thread on windows

mattn:
bash

### DIFF
--- a/plug.vim
+++ b/plug.vim
@@ -761,7 +761,7 @@ endfunction
 function! s:update_impl(pull, force, args) abort
   let args = copy(a:args)
   let threads = (len(args) > 0 && args[-1] =~ '^[1-9][0-9]*$') ?
-                  \ remove(args, -1) : get(g:, 'plug_threads', s:is_win ? 1 : 16)
+                  \ remove(args, -1) : get(g:, 'plug_threads', 16)
 
   let managed = filter(copy(g:plugs), 's:is_managed(v:key)')
   let todo = empty(args) ? filter(managed, '!v:val.frozen || !isdirectory(v:val.dir)') :
@@ -798,9 +798,8 @@ function! s:update_impl(pull, force, args) abort
     echohl None
   endif
 
-  let python = (has('python') || has('python3')) && !s:is_win && !has('win32unix')
-      \ && (!s:nvim || has('vim_starting'))
-  let ruby = has('ruby') && !s:nvim && (v:version >= 703 || v:version == 702 && has('patch374'))
+  let python = (has('python') || has('python3')) && (!s:nvim || has('vim_starting'))
+  let ruby = has('ruby') && !s:nvim && (v:version >= 703 || v:version == 702 && has('patch374')) && !(s:is_win && has('gui_running'))
 
   let s:update = {
     \ 'start':   reltime(),
@@ -1069,7 +1068,6 @@ endfunction
 function! s:update_python()
 let py_exe = has('python') ? 'python' : 'python3'
 execute py_exe "<< EOF"
-""" Due to use of signals this function is POSIX only. """
 import datetime
 import functools
 import os
@@ -1096,6 +1094,7 @@ G_CLONE_OPT = vim.eval('s:clone_opt')
 G_PROGRESS = vim.eval('s:progress_opt(1)')
 G_LOG_PROB = 1.0 / int(vim.eval('s:update.threads'))
 G_STOP = thr.Event()
+G_IS_WIN = vim.eval('s:is_win') == '1'
 
 class PlugError(Exception):
   def __init__(self, msg):
@@ -1110,10 +1109,9 @@ class Action(object):
   INSTALL, UPDATE, ERROR, DONE = ['+', '*', 'x', '-']
 
 class Buffer(object):
-  def __init__(self, lock, num_plugs, is_pull, is_win):
+  def __init__(self, lock, num_plugs, is_pull):
     self.bar = ''
     self.event = 'Updating' if is_pull else 'Installing'
-    self.is_win = is_win
     self.lock = lock
     self.maxy = int(vim.eval('winheight(".")'))
     self.num_plugs = num_plugs
@@ -1141,8 +1139,7 @@ class Buffer(object):
 
     with self.lock:
       vim.command('normal! 2G')
-      if not self.is_win:
-        vim.command('redraw')
+      vim.command('redraw')
 
   def write(self, action, name, lines):
     first, rest = lines[0], lines[1:]
@@ -1224,9 +1221,11 @@ class Command(object):
 
     try:
       tfile = tempfile.NamedTemporaryFile(mode='w+b')
-      self.proc = subprocess.Popen(self.cmd, stdout=tfile,
-                                   stderr=subprocess.STDOUT, shell=True,
-                                   preexec_fn=os.setsid)
+      preexec_fn = not G_IS_WIN and os.setsid or None
+      self.proc = subprocess.Popen(self.cmd, cwd=self.cmd_dir, stdout=tfile,
+                                   stderr=subprocess.STDOUT,
+                                   stdin=subprocess.PIPE, shell=True,
+                                   preexec_fn=preexec_fn)
       thrd = thr.Thread(target=(lambda proc: proc.wait()), args=(self.proc,))
       thrd.start()
 
@@ -1244,7 +1243,7 @@ class Command(object):
 
         if first_line or random.random() < G_LOG_PROB:
           first_line = False
-          line = nonblock_read(tfile.name)
+          line = '' if G_IS_WIN else nonblock_read(tfile.name)
           if line:
             self.callback([line])
 
@@ -1268,7 +1267,10 @@ class Command(object):
   def terminate(self):
     """ Terminate process and cleanup. """
     if self.alive:
-      os.killpg(self.proc.pid, signal.SIGTERM)
+      if G_IS_WIN:
+        os.kill(self.proc.pid, signal.SIGINT)
+      else:
+        os.killpg(self.proc.pid, signal.SIGTERM)
     self.clean()
 
 class Plugin(object):
@@ -1419,10 +1421,9 @@ def main():
   nthreads = int(vim.eval('s:update.threads'))
   plugs = vim.eval('s:update.todo')
   mac_gui = vim.eval('s:mac_gui') == '1'
-  is_win = vim.eval('s:is_win') == '1'
 
   lock = thr.Lock()
-  buf = Buffer(lock, len(plugs), G_PULL, is_win)
+  buf = Buffer(lock, len(plugs), G_PULL)
   buf_q, work_q = queue.Queue(), queue.Queue()
   for work in plugs.items():
     work_q.put(work)
@@ -1479,16 +1480,20 @@ function! s:update_ruby()
 
   def killall pid
     pids = [pid]
-    unless `which pgrep 2> /dev/null`.empty?
-      children = pids
-      until children.empty?
-        children = children.map { |pid|
-          `pgrep -P #{pid}`.lines.map { |l| l.chomp }
-        }.flatten
-        pids += children
+    if /mswin|mingw|bccwin/ =~ RUBY_PLATFORM
+      pids.each { |pid| Process.kill 'INT', pid.to_i rescue nil }
+    else
+      unless `which pgrep 2> /dev/null`.empty?
+        children = pids
+        until children.empty?
+          children = children.map { |pid|
+            `pgrep -P #{pid}`.lines.map { |l| l.chomp }
+          }.flatten
+          pids += children
+        end
       end
+      pids.each { |pid| Process.kill 'TERM', pid.to_i rescue nil }
     end
-    pids.each { |pid| Process.kill 'TERM', pid.to_i rescue nil }
   end
 
   require 'thread'
@@ -1514,7 +1519,7 @@ function! s:update_ruby()
     $curbuf[1] = "#{pull ? 'Updating' : 'Installing'} plugins (#{cnt}/#{tot})"
     $curbuf[2] = '[' + bar.ljust(tot) + ']'
     VIM::command('normal! 2G')
-    VIM::command('redraw') unless iswin
+    VIM::command('redraw')
   }
   where = proc { |name| (1..($curbuf.length)).find { |l| $curbuf[l] =~ /^[-+x*] #{name}:/ } }
   log   = proc { |name, result, type|


mattn:
Please wait. I want to check behavior on gvim.exe

mattn:
Okay, now works fine on gvim.exe

mattn:
`The perl warning` is not issue of vim-plug.

mattn:
#81 is closable?

junegunn:
Thanks!
We disabled the parallel installer by default on Windows due to #81 (also see #191, #165), and looks like you managed to fix the issue, right? I currently don't have a Windows PC to test it, but hopefully @starcraftman will be able to review this as he is the maintainer of the python installer.

plug.vim
```diff
@@ -798,7 +798,7 @@ function! s:update_impl(pull, force, args) abort
     echohl None
   endif
 
-  let python = (has('python') || has('python3')) && !s:is_win && !has('win32unix')
+  let python = (has('python') || has('python3'))
       \ && (!s:nvim || has('vim_starting'))
   let ruby = has('ruby') && !s:nvim && (v:version >= 703 || v:version == 702 && has('patch374'))
```
junegunn: Since your fix only applies to the python installer, we may need to disable the ruby version on windows (which can take precedence over python when both are available).

junegunn:
1. On non-Windows environment, prefer Ruby
2. On Windows, prefer Python and for Python, enable parallel installer by default unless otherwise specified
3. On Windows, if only Ruby is available, use it, but not by default, only when explicitly specified
    - `g:plug_threads` > 1 and not `PlugInstall 1`
    - `PlugInstall N` where N > 1

junegunn:
If that feels too complicated to you, I can take turn and amend the activation conditions.

plug.vim
```diff
@@ -1216,9 +1215,11 @@ class Command(object):
 
     try:
       tfile = tempfile.NamedTemporaryFile(mode='w+b')
+      preexec_fn = not G_IS_WIN and os.setsid or None
       self.proc = subprocess.Popen(self.cmd, cwd=self.cmd_dir, stdout=tfile,
-                                   stderr=subprocess.STDOUT, shell=True,
-                                   preexec_fn=os.setsid)
+                                   stderr=subprocess.STDOUT,
+                                   stdin=subprocess.PIPE, shell=True,
```
starcraftman: Why is stdin connected at all? I don't think it is being used, so can be deleted.

plug.vim
```diff
@@ -1260,7 +1261,10 @@ class Command(object):
   def terminate(self):
     """ Terminate process and cleanup. """
     if self.alive:
-      os.killpg(self.proc.pid, signal.SIGTERM)
+      if not G_IS_WIN:
```
starcraftman: I try and avoid negations with if's when I can just invert which block comes first. Reads more naturally. Can you just rearrange this so that it is

````python
if G_IS_WIN:
  os.kill(self.proc.pid, signal.SIGINT)
else:
  os.killpg(self.proc.pid, signal.SIGTERM)
```

mattn:
> On Windows, if only Ruby is available, use it, but not by default, only when explicitly specified

ruby implementation may not work on windows. Because this use `/dev/null` or some unix specific feature. So I sugguest disable ruby on windows.

plug.vim
mattn: gvim.exe doesn't have STDIN. so it MUST NOT use stdin directly. When in default or specified None, it make error.

plug.vim
```diff
@@ -1394,9 +1398,13 @@ def esc(name):
...
-  fread = os.open(fname, os.O_RDONLY | os.O_NONBLOCK)
-  buf = os.read(fread, 100000).decode('utf-8', 'replace')
-  os.close(fread)
+  try:
+    nonblock_flag = not G_IS_WIN and os.O_NONBLOCK or 0
+    fread = os.open(fname, os.O_RDONLY | nonblock_flag)
+    buf = os.read(fread, 100000).decode('utf-8', 'replace')
+    os.close(fread)
+  except:
+    return ''
```
starcraftman: Doesn't quite seem right to me, since now it violates my description of returning the last line of the file to be read. How about instead, we simply only make the call if we aren't on Windows shell.

- Undo the changes to nonblock_read function.
- Then change line 1339 to:
```python
          line = '' if G_IS_WIN else nonblock_read(tfile.name)
```

plug.vim
starcraftman: @mattn Are you sure about that? My understanding is that it should inherit the parents file descriptors which should be set to null, since stdin never gets set. When I took it out, the python installer works fine with my GVim. Seems weird.

mattn:
Added condition to choice python/ruby/vim that @junegunn says. And I added some fix for ruby implementation. I confirmed ruby implementation is working.

mattn:
Ah, sorry. please wait if you want to merge now.

plug.vim
```diff
@@ -1236,7 +1258,7 @@ class Command(object):
 
         if first_line or random.random() < G_LOG_PROB:
           first_line = False
-          line = nonblock_read(tfile.name)
+		  line = '' if G_IS_WIN else nonblock_read(tfile.name)
```
starcraftman: Whitespace a bit off. :)

mattn: Yes, I just notice it.

starcraftman:
@mattn Thanks for the PR by the way, surely will be appreciated by the Windows people. I think I got tunnel visioned on getting parity for my linux python installer, when it works just fine if I had dropped the nonblock bit.

That bit with all those if's and logical &'s re python/ruby is really getting out of hand. Maybe we can refactor that somehow at a later date.

junegunn:
@mattn Thanks for the follow up. If you can fix Ruby installer as well that's really great and then we don't have to follow the complex activation condition I suggested above. We can simplify the rule to:

- Activate parallel installer by default
- Prefer ruby installer both on Windows and non-Windows platform

mattn:
```diff
        def _get_handles(self, stdin, stdout, stderr):
            """Construct and return tuple with IO objects:
            p2cread, p2cwrite, c2pread, c2pwrite, errread, errwrite
            """
            to_close = set()
            if stdin is None and stdout is None and stderr is None:
                return (None, None, None, None, None, None), to_close

            p2cread, p2cwrite = None, None
            c2pread, c2pwrite = None, None
            errread, errwrite = None, None

            if stdin is None:
                p2cread = _subprocess.GetStdHandle(_subprocess.STD_INPUT_HANDLE)
                if p2cread is None:
                    p2cread, _ = _subprocess.CreatePipe(None, 0)
            elif stdin == PIPE:
                p2cread, p2cwrite = _subprocess.CreatePipe(None, 0)
            elif isinstance(stdin, int):
                p2cread = msvcrt.get_osfhandle(stdin)
            else:
                # Assuming file-like object
                p2cread = msvcrt.get_osfhandle(stdin.fileno())
```
This is a part of subprocess.py that is python2 codes installed by msys2 installer.
gvim is GUI application, so the handle of stdin doesn't have same privileges with stdin.

mattn:
oops.

mattn:
> Activate parallel installer by default
> Prefer ruby installer both on Windows and non-Windows platform

okay, then reverted making condition. it should be your expected.

starcraftman:
@mattn Ok then well leave it connecting to PIPE, just seems weird I can't reproduce. I checked my python code is same in subprocess, installed with activestate standalone using cmd.exe though not msys.

junegunn:
@mattn Looks good. But the changes you made in Ruby installer don't seem to be related to the problem described #81, so how is the issue fixed? Is it fixed in the recent versions of GVim?

mattn:
This PR fixes:

* Enable mult-thread on windowos. (Also fixes issue of gvim.exe)
* Fixes some problem on ruby implementation

starcraftman:
@mattn You seem to have rebased too far and cloberred some of the existing master history into your branch. Please fix and then squash down your 3 commits to a single commit. After that, I don't see any problem doing a merge for this PR. Thanks again.

junegunn:
@mattn Hmm, are you suggesting that we are still going to have some issues (#81) with Ruby installer? What about Python installer? #81 was the reason we decided to disable parallel installers on Windows and if it's not yet fixed, I'm not sure about reverting the decision.

@starcraftman Can you check if #81 is fixed for both installer?

mattn:
@starcraftman okay, rebased.

@junegunn As you can see in my screenshot on above, progress is displayed because i removed condition to disable calling `redraw`.

junegunn:
@mattn I see, the `unless iswin` part was added [long long time ago](https://github.com/junegunn/vim-plug/commit/8bfcf21f2269c484ffbd0c80a42f380831669f9a), and may not be relevant anymore. It's pretty astonishing that simply removing the old, forgotten piece of code fixes this long-standing issue.

junegunn:
@mattn As @starcraftman requested, can you squash the commits into one? We don't need to have 3 separate commits for this. Thanks!

mattn:
done. thank you

starcraftman:
@junegunn I tested the ruby installer and it hangs for me on Windows. The python installer though seems to function perfectly. My setup is GVim + cmd.exe and python/ruby installed separately. @mattn is the ruby installer working for you fine? Are you only testing against msys2?

mattn:
> is the ruby installer working for you fine? Are you only testing against msys2?

Yes, I tested ruby2.2 on mingw64/msys2.

starcraftman:
@mattn Interesting. Try doing the following Windows setup, I'm curious if it hangs for you:
1. Download standard [gvim](http://www.vim.org/download.php#pc)
1. Replace that gvim, with the one from here > http://wyw.dcweb.cn/#download
1. Install [ruby](http://rubyinstaller.org/downloads/) 2.0.0 32bit

That's how we instruct people to setup GVim standalone for ruby. Seems to hang on my system.

mattn:
Sorry, I have a principle that don't use vim my build.

What version of ruby do you use? I'm using vim of master/HEAD, ruby2.2.0.

starcraftman:
@mattn Well ok, if you insist on building it yourself. The installer has to work with the method we tell people, and msys is only one of the two ways we support.

Here is the version of the vim I am using. The ruby version I used was 2.0.0-p647, 32 bit.
```vim
VIM - Vi IMproved 7.4 (2013 Aug 10, compiled Nov 25 2015 10:21:43)
MS-Windows 32-bit GUI version with OLE support
Included patches: 1-941
Compiled by Yongwei@Shanghai
Big version with GUI.  Features included (+) or not (-):
+acl                +cryptv             +file_in_path       +listcmds           +persistent_undo    +statusline         +visual
+arabic             +cscope             +find_in_path       +localmap           -postscript         -sun_workshop       +visualextra
+autocmd            +cursorbind         +float              +lua/dyn            +printer            +syntax             +viminfo
+balloon_eval       +cursorshape        +folding            +menu               -profile            +tag_binary         +vreplace
+browse             +dialog_con_gui     -footer             +mksession          +python/dyn         +tag_old_static     +wildignore
++builtin_terms     +diff               +gettext/dyn        +modify_fname       +python3/dyn        -tag_any_white      +wildmenu
+byte_offset        +digraphs           -hangul_input       +mouse              +quickfix           +tcl/dyn            +windows
+cindent            +directx            +iconv/dyn          +mouseshape         +reltime            -tgetent            +writebackup
+clientserver       -dnd                +insert_expand      +multi_byte_ime/dyn +rightleft          -termresponse       -xfontset
+clipboard          -ebcdic             +jumplist           +multi_lang         +ruby/dyn           +textobjects        -xim
+cmdline_compl      +emacs_tags         +keymap             -mzscheme           +scrollbind         +title              -xterm_save
+cmdline_hist       +eval               +langmap            +netbeans_intg      +signs              +toolbar            +xpm_w32
+cmdline_info       +ex_extra           +libcall            +ole                +smartindent        +user_commands
+comments           +extra_search       +linebreak          +path_extra         -sniff              +vertsplit
+conceal            +farsi              +lispindent         +perl/dyn           +startuptime        +virtualedit
   system vimrc file: "$VIM\vimrc"
     user vimrc file: "$HOME\_vimrc"
 2nd user vimrc file: "$HOME\vimfiles\vimrc"
 3rd user vimrc file: "$VIM\_vimrc"
      user exrc file: "$HOME\_exrc"
  2nd user exrc file: "$VIM\_exrc"
  system gvimrc file: "$VIM\gvimrc"
    user gvimrc file: "$HOME\_gvimrc"
2nd user gvimrc file: "$HOME\vimfiles\gvimrc"
3rd user gvimrc file: "$VIM\_gvimrc"
    system menu file: "$VIMRUNTIME\menu.vim"
Compilation: cl -c /W3 /nologo  -I. -Iproto -DHAVE_PATHDEF -DWIN32   -DFEAT_CSCOPE -DFEAT_NETBEANS_INTG   -DFEAT_XPM_W32   -DWINVER=0x0501 -
D_WIN32_WINNT=0x0501  /Fo.\ObjGXOULYHTRi386/ /Ox /GL -DNDEBUG  /Zl /MT -DFEAT_OLE -DFEAT_MBYTE_IME -DDYNAMIC_IME -DGLOBAL_IME -DFEAT_MBYTE -
DFEAT_GUI_W32 -DFEAT_DIRECTX -DDYNAMIC_DIRECTX -DDYNAMIC_ICONV -DDYNAMIC_GETTEXT -DFEAT_TCL -DDYNAMIC_TCL -DDYNAMIC_TCL_DLL=\"tcl85.dll\" -D
DYNAMIC_TCL_VER=\"8.5\" -DFEAT_LUA -DDYNAMIC_LUA -DDYNAMIC_LUA_DLL=\"lua51.dll\" -DFEAT_PYTHON -DDYNAMIC_PYTHON -DDYNAMIC_PYTHON_DLL=\"pytho
n27.dll\" -DFEAT_PYTHON3 -DDYNAMIC_PYTHON3 -DDYNAMIC_PYTHON3_DLL=\"python32.dll\" -DFEAT_PERL -DDYNAMIC_PERL -DDYNAMIC_PERL_DLL=\"perl514.dl
l\" -DFEAT_RUBY -DDYNAMIC_RUBY -DDYNAMIC_RUBY_VER=20 -DDYNAMIC_RUBY_DLL=\"msvcrt-ruby200.dll\" -DFEAT_BIG /Fd.\ObjGXOULYHTRi386/ /Zi
Linking: link /RELEASE /nologo /subsystem:windows,5.01 /LTCG:STATUS oldnames.lib kernel32.lib advapi32.lib shell32.lib gdi32.lib  comdlg32.l
ib ole32.lib uuid.lib /machine:i386 gdi32.lib version.lib   winspool.lib comctl32.lib advapi32.lib shell32.lib  /machine:i386 /nodefaultlib
libcmt.lib oleaut32.lib user32.lib   /nodefaultlib:lua51.lib   /nodefaultlib:python27.lib /nodefaultlib:python32.lib   "C:\Tcl\lib\tclstub85
.lib" WSock32.lib C:\Libraries\xpm\lib\libXpm.lib /PDB:gvim.pdb -debug                                                                     
```

junegunn:
@starcraftman It's okay that we change the instruction if there are better alternatives. But if we can't figure out a way to make Ruby installer reliably work on Windows GVim in any way, we'll have to disable Ruby installer on it (`... && !s:is_win || !has('gui_running')`)

starcraftman:
@junegunn Oh yes, I'm fine with either changing instructions or disabling it. I'm just saying on what I consider a fairly standard GVim setup it seems to hang for me. If the ruby installer is to be enabled on Windows, seems at least to me this setup should work.

junegunn:
@starcraftman Yeah that's why we didn't enable parallel installer by default. It's unstable on Windows GVim. In order to keep perfect backward compatibility we'll have to embrace the complex activation logic above but I don't like it either. So I'm suggesting completely turning off Ruby installer on Windows GVim, while keeping it alive for non-gui Windows vim as it's known to work well, so we limit the number of hypothetical users that are affected by the change. We also need to update the wiki page to encourage users to install python-enabled GVim instead on Windows.

starcraftman:
@junegunn Alright, I'll do a bit more testing tomorrow making sure the python setups ok. I agree the docs need updating, we'll look at that soon.

starcraftman:
Just a note on what I've tested so far, checked boxes mean working. Unless otherwise specified, everything was win32.

- [x] GVim + Activestate python 2.7.6 + cmd.exe + msysgit for git
- [ ] GVim + Ruby 2.0.0 + cmd.exe + msysgit for git
  - Hangs after popping up a bunch of cmd.exe prompts, appears all threads crash but main never realizes.

For msys2 I used the pacman versions, so `pacman -S vim ruby python git diff-utils`
- [ ] msys2 + vim/python installer
  - Gives segv, seems to have to do with Plugin.repo_uri() method, which is peculiar. Maybe some escaping issue?
- [x] msys2 + vim/ruby installer 

I think I want to check babun and cygwin too, though not sure how well that will do. Just getting those first 4 test cases would be good enough.

junegunn:
@starcraftman Thank you very much. Hmm, it's unfortunate that python installer on msys2 crashes as well. So currently, considering that we don't find a fix, we need to apply the following conditions, right?

- ruby: `&& !(s:is_win && has('gui_running'))`
- python: `&& !(s:is_win && !has('gui_running'))`

starcraftman:
@junegunn It is a strange crash, python installer seems to fail to run basic shell commands but only when the change directory arg is given to Popen. I have a simple fix that I will test and upload in a bit. If works, then we can just rebase this PR and have python working on both cases on merge.

The logic looks right to me.

junegunn:
@starcraftman Cool, thanks for looking into this :) I wish it works and we could simplify the conditions.

starcraftman:
@junegunn No problem. I've tested my tiny patch and it appears to work. Means only remaining broken installers is msys + ruby. I'll go ahead and merge my patch, then @mattn can rebase and add the ruby gui check.

starcraftman:
@mattn To summarize, please do the following:
- Rebase to master.
- Change line around 800 that starts  `  let ruby = ... ` to
```viml
  let ruby = has('ruby') && !s:nvim && (v:version >= 703 || v:version == 702 && has('patch374')) && !(s:is_win && has('gui_running'))
```
- We can remove this comment: `""" Due to use of signals this function is POSIX only. """`.

I think after that, we are good to go.

mattn:
okay, done.

junegunn:
@mattn Thanks again :)

@starcraftman The python tests are failing, is it expected?

starcraftman:
@junegunn I made a PR directly onto @mattn 's branch, it was just a stray attribute access that no longer exists left in during rebase merge. We should be good to go once he updates the branch.

starcraftman:
@mattn Just curious, we've removed the win32unix check for msys2, did you check how vim-plug fared with cygwin and/or babun? I'm mainly curious, I guess we can deal with any bugs down the road, even manually check for babun/cygwin at startup if they give too much hassle. Otherwise, just squash the commits down into the first and I think we can pull.

starcraftman:
Just tried it myself inside a virtual Windows 10 machine. Seems that babun + vim + python is working. I'm happy then. 

@junegunn You ok with merging this once the commits squashed?

junegunn:
Yep, looks good to me. Thanks.

junegunn:
I squashed the commits myself and merged to master. Let me know if there's any problem. Many thanks to you both.

mattn:
great! :+1:

PR #357, opened 2015-12-13
Fix #353: Not Working On Windows

### LINKED ISSUE #353: Not Working on Windows
It used to work well, but after the recent update, now whenever I call `PlugUpdate/PlugInstall`, I get errors like:

`fatal: could not create work tree dir 'D:\vimfiles\plugged\nerdtree" '.: Invalid argument`

starcraftman:
* First fix was only partial, allow drive letter changes.

### DIFF
--- a/plug.vim
+++ b/plug.vim
@@ -1168,10 +1168,12 @@ class Buffer(object):
       pass
 
 class Command(object):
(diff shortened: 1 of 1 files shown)


starcraftman:
Should be it for a while, still need to look at that cmder issue when I can.

PR #364, opened 2015-12-19
Add git installation, and use vim-plug to manage vim-plug

frtmelody:
Add git installation, and use vim-plug to manage vim-plug


junegunn:
Thanks but we decided not to do this. See https://github.com/junegunn/vim-plug/wiki/faq#shouldnt-vim-plug-update-itself-on-plugupdate-like-vundle

PR #386, opened 2016-01-24
Only highlight vim-plug buffers if syntax highlighting is enabled

srstevenson:
Previously, syntax highlighting was turned on unconditionally in the buffers opened by commands such as `:PlugUpdate`. With this commit, highlighting is not applied if the user has disabled syntax highlighting
by calling `syntax off` after `call plug#end()` in their `vimrc`.


junegunn:
Looks good to me. Thanks.

Issue #388, opened 2016-01-27
Please document basic commands in vim help file.


chrisbra:
As per the title. Thanks.

junegunn:
This was discussed before in #234, but we were unable to reach a satisfactory conclusion as `plug.vim` was designed to be distributed as a standalone vim script file. As noted in the issue, I can easily make a help file, but we'll have to decide where to put it.

chrisbra:
Hm, I don't follow what would be so terrible bad about putting the documention below `~/.vim/doc` After all, that's what the directory is for. But perhaps you want to distribute the plugin additionally as vimball for those who want documentation. I would then just download the .vba file and be done with it.

junegunn:
What I meant by "where to put it" in the above comment was about the position of the file in this git repo. I was worried about this repo having `doc` directory as it can give the wrong impression that it is "pathogen-compatible". Anyway, I have no objection on the use of `~/.vim/doc`. I was just saying in the previous discussion that vim-plug should not "automagically" put things into the directory without the user's consent.

chrisbra:
okay, I understand. Therefore I made it self-extractable #390

starcraftman:
Uh, this again. I dunno, in #234 I argued for it, I've since changed my mind. I don't see much point for the fuss we will need (like stuffing the doc into plug.vim as you did and then extracting).

Let me ask you @chrisbra , what questions exactly would you have done `:h vim-plug` for? Apart from the information needed to do initial startup/conversion to vim-plug I find vim-plug to be very intuitively named with regards to commands/options. I'd argue since most people will stumble upon vim-plug via their browser (i.e. r/vim, github, so on) they will read the README while converting. I suspect only a handful of less used options/flags would ever be consulted.

As to location, I git manage my vim folder (as quite a few do) and would not like it in `~/.vim/doc`. Would require me ignoring another folder. More importantly, vim-plug has permission to write to the path passed to `plug#begin()`, it shouldn't write elsewhere.

chrisbra:
@starcraftman 
1) I wanted find out, what arguments are allowed for the various `:Plug` commands. #389 
2) Supposedly, there are people, who don't have internet all the time available. And I don't like having to find the README.md file manually and reading it. That is much more trouble than saying `:h plugin.txt`
3) I don't understand. ~/.vim/doc is exactly for that kind of thing. plug.vim resides in ~/.vim/autoload so the documentation should be in ~/.vim/doc
[...]

PR #390, opened 2016-01-27
generate documentation from script

chrisbra:
#388

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,13 @@ $uri = 'https://raw.githubusercontent.com/junegunn/vim-plug/master/plug.vim'
 (New-Object Net.WebClient).DownloadFile($uri, $ExecutionContext.SessionState.Path.GetUnresolvedProviderPathFromPSPath("~\vimfiles\autoload\plug.vim"))
 ```
 
+#### Install the documentation (optionally)
+
+```sh
+vim -u NONE -N -S plug_extract_doc.vim
+```
+Read the help at `:h plug.txt`
+
 ### Usage
 
 Add a vim-plug section to your `~/.vimrc` (or `~/.config/nvim/init.vim` for Neovim):
--- a/plug.vim
+++ b/plug.vim
@@ -2117,4 +2117,253 @@ endif
 
 let &cpo = s:cpo_save
 unlet s:cpo_save
+finish
+*plug.txt*	A Minimal Plugin Manager
 
+PLUG - TABLE OF CONTENTS                                         *plug* *plug-toc*
+==============================================================================
+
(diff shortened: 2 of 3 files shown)


chrisbra:
don't understand why travis failed, as I did not change the plugin at all.

junegunn:
Well, the failure is due to the test script appending some code to plug.vim for clearing the internal data structure during the test. Yeah, It's hacky and can have some improvement, but I just didn't want to add some API just for test.

Anyway, thanks for putting this together but I don't think I'm in favor of this approach.
- First of all, the markdown document you attached is not up-to-date.
- `plug.vim` is already pretty long, and we're adding 200 more lines here.
- It's basically two files of completely different types concatenated and it doesn't feel like an elegant solution. The merged file is not editor-friendly and will be difficult to maintain.
- The help file is not updated on `PlugUpgrade` which can be confusing.
- If we're going to tell the user to use an extra script for installing the documentation, what's the point of having the text included in plug.vim? I would just separately provide plug.txt in this repository and make `PlugUpgrade` or some other command handle that.

We'll continue the discussion, but this is not the approach I have in mind.

chrisbra:
I don't agree with alll those points, but oh well..

junegunn:
So you think vimscript file concatenated with a markdown file is an elegant solution? Hmm, we'll have to agree to disagree then.

chrisbra:
No, I think when distributing a single plugin.vim file, this is the only possibility to keep documentation. But nevertheless my approach did not work, since I introduced a new file, which would not be available after the typical curl... installation. 
BTW: it is not that uncommon, to distribute documentation included with code. perl pod does exactly this.

junegunn:
Okay thanks for the clarification. Looks like I somehow misinterpreted your words, sorry about that. Yes, perl pod is a good example but we both know that the situation here is different from that. If we don't have to provide help in vimdoc format, we can do similar stuff with the header comment as I mentioned in #388.

vheon:
If I can give my 2 cents:

- Ship the vimdoc documentation as the header of the plug.vim file surrounded by some delimiter like `STARTDOC` `ENDDOC`.
- Instead of making a separate vim script add a command like `PlugGenerateDoc` to create the `~/.vim/doc/plug.txt` file and generate the helptags
- On `PlugUpgrade` simply rerun the `PlugGenerateDoc` automatically if the file is already there (i.e. the user already allow vim-plug to write the documentation)

I find myself from time to time to want the vim-plug documentation and I would prefer to have it right inside vim instead of going to the browser (not that is that of a big deal anyway, but oh well...)

PR #395, opened 2016-01-29
Fix #394: Avoid problem happening.

### LINKED ISSUE #394: weird neovim key not present in dictionary errors
If I run `:PlugUpdate` and press `D` or `S` neovim starts throwing out errors about keys not being present in dictionaries. I'm not sure what the problem is. I don't think it has to do with my neovim configuration because it still happens with a init.vim with only one plugin.

screen shot 2016-01-29 at 10 31 50 am

It continues and continues and continues with more keys not found and stuff.

starcraftman:
This is largely a response to fix: #394. 

Maps are only bound on prepare when function is guaranteed synchronous, else delay until finish of update/install to be done in s:finish(). Avoids the accidental neovim breakage experienced in that issue.

Also, small typo with respect to the platform test, print doesn't need brackets for python 2.

TODO:
- [x] Don't write pending updates if no commit/tag opts.
- [x] Resolve buffer reuse.

### DIFF
--- a/plug.vim
+++ b/plug.vim
@@ -620,25 +620,32 @@ function! s:switch_out(...)
   endif
 endfunction
 
-function! s:prepare()
+function! s:finish_bindings()
+  nnoremap <silent> <buffer> R  :silent! call <SID>retry()<cr>
+  nnoremap <silent> <buffer> D  :PlugDiff<cr>
+  nnoremap <silent> <buffer> S  :PlugStatus<cr>
+  nnoremap <silent> <buffer> U  :call <SID>status_update()<cr>
+  xnoremap <silent> <buffer> U  :call <SID>status_update()<cr>
+  nnoremap <silent> <buffer> ]] :silent! call <SID>section('')<cr>
+  nnoremap <silent> <buffer> [[ :silent! call <SID>section('b')<cr>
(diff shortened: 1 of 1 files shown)


junegunn:
Thank you. The remaining problem is as we continuously reuse the buffer when it's open, it fails when you do `PlugUpdate` again while the window is open and press `D`. We can either 1. unmap those mappings, 2. or change the code not to reuse the buffer but simply wipes it out and create another one (`enew`?).

starcraftman:
@junegunn Good point, I'll add that to my todo.

starcraftman:
@junegunn Anything else? I went with simply destroying it every time by invoking q.

junegunn:
How about enew?

```vim
  if s:switch_in()
    enew
  else
    call s:new_window()
  endif
```

This clears the buffer (bufhidden=wipe) and its maps but reuses the window, so the size of the window is retained if the user previously resized it for some reason.

junegunn:
After a second thought, I think `q` (which is essentially `bd`) is safer. Since a crazy user might `tab split` the buffer :)

junegunn:
:+1:

starcraftman:
No problem.

PR #399, opened 2016-02-01
Fix #397: clarify in docs that GitHub is the default plugin source

### LINKED ISSUE #397: clarify in docs that GitHub is the default plugin source
There seems to be nothing in the docs clarifying that some plugin syntax 'blah/someplug' is GitHub. There shouldn't be any assumption that everything lives on GitHub, even though in practice that's the case today.

I'd like to see it stated explicitly that GitHub is the default for that syntax and also see examples of alternate syntax for other plugin source repositories.

starcraftman:
Made changes discussed in #397. Also, put a copy of the Plug options table in the plug.vim file. I think this was suggested previously by a few people that just wanted a quick reminder, doesn't take up too much space.

@junegunn Comments?

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,8 +63,12 @@ Add a vim-plug section to your `~/.vimrc` (or `~/.config/nvim/init.vim` for Neov
 call plug#begin('~/.vim/plugged')
 
 " Make sure you use single quotes
-Plug 'junegunn/seoul256.vim'
+
+" Shorthand notation; fetches https://github.com/junegunn/vim-easy-align
 Plug 'junegunn/vim-easy-align'
+
+" Any valid git URL is allowed
+Plug 'https://github.com/junegunn/vim-github-dashboard.git'
 
 " Group dependencies, vim-snippets depends on ultisnips
 Plug 'SirVer/ultisnips' | Plug 'honza/vim-snippets'
@@ -72,9 +76,6 @@ Plug 'SirVer/ultisnips' | Plug 'honza/vim-snippets'
 " On-demand loading
 Plug 'scrooloose/nerdtree', { 'on':  'NERDTreeToggle' }
 Plug 'tpope/vim-fireplace', { 'for': 'clojure' }
-
-" Using git URL
-Plug 'https://github.com/junegunn/vim-github-dashboard.git'
 
 " Using a non-master branch
 Plug 'rdnetto/YCM-Generator', { 'branch': 'stable' }
--- a/plug.vim
+++ b/plug.vim
@@ -11,8 +11,12 @@
 "   call plug#begin('~/.vim/plugged')
 "
 "   " Make sure you use single quotes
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -73,7 +74,7 @@ Plug 'SirVer/ultisnips' | Plug 'honza/vim-snippets'
 Plug 'scrooloose/nerdtree', { 'on':  'NERDTreeToggle' }
 Plug 'tpope/vim-fireplace', { 'for': 'clojure' }
 
-" Using git URL
+" Any valid git URL is allowed
 Plug 'https://github.com/junegunn/vim-github-dashboard.git'
```
starcraftman: I'm open to a non-Github plugin if anyone has a good suggestion...

@vheon thoughts?

vheon: The only one that I know was `http://git.wincent.com/command-t.git` but even this one has a mirror on Github.

(did you really wanted to ping me? or it was for junegunn? :stuck_out_tongue_closed_eyes: )

starcraftman: @vheon Ya I meant you.I think junegunn's drawing a blank like me. Ah well, I guess we'll just use stick to the full url github one that is present.

junegunn: How about if we move line 77 and 78 upward just after the first shorthand notation example? So it goes, "This is the shorthand notation, but you can still write a full git url".

plug.vim
```diff
@@ -40,7 +41,21 @@
...
+"| `branch`/`tag`/`commit` | Branch/tag/commit of the repository to use       |
+"| `rtp`                   | Subdirectory that contains Vim plugin            |
+"| `dir`                   | Custom directory for the plugin                  |
+"| `as`                    | Use different name for the plugin                |
+"| `do`                    | Post-update hook (string or funcref)             |
+"| `on`                    | On-demand loading: Commands or `<Plug>`-mappings |
+"| `for`                   | On-demand loading: File types                    |
+"| `frozen`                | Do not update unless explicitly specified        |
+"
+" More Information: https://github.com/junegunn/vim-plug
```
junegunn: Can we use "Plug options:" and "More information:"? No reason, uncapitalized version just looks better to my eyes.

plug.vim
```diff
@@ -21,7 +22,7 @@
 "   Plug 'scrooloose/nerdtree', { 'on':  'NERDTreeToggle' }
 "   Plug 'tpope/vim-fireplace', { 'for': 'clojure' }
 "
-"   " Using git URL
+"   " Any valid git URL is allowed
```
junegunn: Likewise, let's move this to line 17.

README.md
starcraftman: Agreed, done.

starcraftman:
@junegunn Anything else or merge?

junegunn:
Looking great. Thanks!

Issue #367, opened 2015-12-23
Have vim-plug manage itself (similar to vundle)


hallzy:
I think it would be nice to have the repo restructured so that it is easy to use vim-plug to manage vim-plug. It would be useful for pulling updates that are made to this repo, and it is the same thing that vundle recommends doing.

The changes would not be complex, in fact, there is already a fork of this repo that has this set up: see this [fork](https://github.com/frtmelody/vim-plug)

hallzy:
Sorry, I just saw #364 saying that you have decided not to do this.

starcraftman:
@hallzy Its okay, you probably won't be the last to suggest it. :)

PR #417, opened 2016-02-19
Adding Updating instructions.

hiberno:


starcraftman:
@hiberno Hello, I'm going to have to decline the PR.

This has been raised before however (i.e. #367, #364), it is intentional that we don't manage vim-plug with vim-plug, a clear discussion was on issue #240.

We already have a similar tip in our [faq](https://github.com/junegunn/vim-plug/wiki/faq#shouldnt-vim-plug-update-itself-on-plugupdate-like-vundle). I don't think we've any interest in promoting the idea more than that.

junegunn:
@starcraftman Thanks for the explanation

@hiberno I think `PlugUpdate | PlugUpgrade` is the simplest and the most straightforward solution, thus updated the wiki page accordingly.

https://github.com/junegunn/vim-plug/wiki/faq#shouldnt-vim-plug-update-itself-on-plugupdate-like-vundle

hiberno:
@starcraftman Oh. I should've researched better. Sorry for the noise. 

@junegunn Thanks for the hint.

Issue #278, opened 2015-09-07
[neovim] PlugUpdate/Install as start commands don't show any UI 


cHoco:
Hello, a couple of days ago I updated vim-plug to the latest version.
Today I tried to run the usual `nvim +PlugUpdate` and even though the
update is performed correctly the UI isn't updated/shown until the 
process is finished. The same happens when running PlugInstall.

Everything works correctly if I execute both commands when neovim
has already started. 

The issue seems to have been introduced with https://github.com/junegunn/vim-plug/pull/227 but I don't
understand if it's something expected with that change, or I'm experiencing a bug.

vheon:
> Today I tried to run the usual vim +PlugUpdate and even though the

> Everything works correctly if I execute both commands when neovim

So its vim or neovim??

cHoco:
Neovim, sorry for the typo :(

junegunn:
Running the latest neovim?

cHoco:
Yeah, latest neovim version + latest neovim python-client.
I can reproduce the bug both on latest OS X version and Ubuntu 14.04,
even with a minimal vimrc:
```
call plug#begin('~/.vim/plugged')

" some plugins just so it takes sometime to fetch them
Plug 'tpope/vim-repeat'
Plug 'tpope/vim-surround'
Plug 'tpope/vim-unimpaired'

call plug#end()
``` 
To be clear vim isn't affected by this problem.

junegunn:
Could be a neovim issue. It worked for me, until I updated neovim this morning (`brew reinstall neovim --HEAD`). Too bad I forgot to write down the version before upgrading it :(

cHoco:
Just did a rapid check and found out that this is the commit in neovim causing the bug in vim-plug https://github.com/neovim/neovim/commit/7475c1c0f795912ea5a4613a23c4ef5a1d88c24d and the respective PR https://github.com/neovim/neovim/pull/2846

The first comment to the first diff is probably the reason why it doesn't work.

cHoco:
And here's a diff for the possible fix, if I understood vim-plug code and the new constraints of the python-client correctly:

```diff
diff --git a/plug.vim b/plug.vim
index 9eb11fc..99cde71 100644
--- a/plug.vim
+++ b/plug.vim
@@ -1121,7 +1121,11 @@ class Buffer(object):
           lnum = 3
       else:
         lnum = 3
-      curbuf.append(msg, lnum)
+
+      if lnum == 3:
[...]

PR #419, opened 2016-02-21
Avoid multiple `syntax enable` during Vim startup

itspriddle:
If one runs `vim +PlugUpdate` there are two calls to `syntax enable`; one from `plug#end()` in your vimrc and another from the `PlugUpdate` command itself. It seems that when you use `+PlugUpdate` via the CLI that *both* calls to `plug#end()` happen while `has('vim_starting')` is true. This causes a minor issue where syntax highlighting (for the plug.vim itself) is lost when `PlugUpdate` finishes running.

There is a gif I found in an [unrelated issue](https://github.com/junegunn/vim-plug/issues/278#issuecomment-138683431) that shows this in action.

With this patch syntax highlighting is not lost when `PlugUpdate` finishes.

### DIFF
--- a/plug.vim
+++ b/plug.vim
@@ -248,7 +248,9 @@ function! plug#end()
   call s:reorg_rtp()
   filetype plugin indent on
   if has('vim_starting')
-    syntax enable
+    if has('syntax') && !exists('g:syntax_on')
+      syntax enable
+    end
   else
     call s:reload()
   endif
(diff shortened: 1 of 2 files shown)


junegunn:
Looks good, thanks. Can you update the failing assertion as well? You can run the whole test with `test/run` script.

itspriddle:
Yep, should be good. Let me know if not, I havent used vader before :smile:

junegunn:
Thanks!

PR #447, opened 2016-03-17
Added 'via' option, to allow to switch between https and ssh

rperce:
Git only allows ssh push if you have two-factor authentication turned on. I do, which means any plugins I write that I'm using can't be updated from inside `plugged`. Obviously, I could just change the url to `git@github.com:user/repo` instead of just `user/repo`, but that felt slightly hacky. So I added a `via` option, so my plug can look like
    
    Plug 'user/repo', { 'via': 'ssh' }

### DIFF
--- a/plug.vim
+++ b/plug.vim
@@ -54,6 +54,7 @@
 "| `on`                    | On-demand loading: Commands or `<Plug>`-mappings |
 "| `for`                   | On-demand loading: File types                    |
 "| `frozen`                | Do not update unless explicitly specified        |
+"| `via`                   | Use either https or ssh to clone                 |
 "
 " More information: https://github.com/junegunn/vim-plug
 "
@@ -484,7 +485,7 @@ function! s:Plug(repo, ...)
     let repo = s:trim(a:repo)
     let opts = a:0 == 1 ? s:parse_options(a:1) : s:base_spec
     let name = get(opts, 'as', fnamemodify(repo, ':t:s?\.git$??'))
-    let spec = extend(s:infer_properties(name, repo), opts)
+    let spec = extend(s:infer_properties(name, repo, opts), opts)
     if !has_key(g:plugs, name)
       call add(g:plugs_order, name)
     endif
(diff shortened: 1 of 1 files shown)


rperce:
Note: I saw #62, but I didn't really want to switch _everything_ to ssh with `g:plug_url_format` (https _is_ recommended, after all). That said, I totally understand if you'd rather not add this. It would let you add more protocols in the future if github ever has more protocols, I guess.

junegunn:
Thanks for the suggestion, but vim-plug labels itself as "a minimalist plugin manager" and the maintainers are trying hard to keep the number of options as small as possible. The basic principle is that we don't add an option or a feature if it's not necessary, even if it's trivial to implement. I don't think using `git@github.com:user/repo` is hacky. On the contrary, I've noticed that some users don't feel good about just using `user/repo` and prefer to write down the whole URL. It's a preference thing. One problem with the new option you suggested is that it's not compatible with `g:plug_url_format`. Your implementation gives priority to `g:plug_url_format`, but some may not agree with the decision. Which one should we prefer? It's not clear. It's an example of added complexity.

> I didn't really want to switch everything to ssh with g:plug_url_format

The variable only applies the subsequent Plug commands that follow. I guess it's not very straightforward but it allows you to write something like the following:

```vim
" Default URL pattern
Plug 'foo/bar'
Plug 'foo/baz'

" Plugins on bitbucket
let g:plug_url_format = 'https://bitbucket.org/%s.git'
Plug 'hello/world'
Plug 'goodbye/world'

" SSH + GitHub
let g:plug_url_format = 'git@github.com:%s.git'
Plug 'apple/juice'
Plug 'banana/sundae'

unlet g:plug_url_format
```

Closing as wontfix, but thanks again for looking into it.

junegunn:
I guess we need to update [the description of the variable](https://github.com/junegunn/vim-plug#global-options).

rperce:
Oh, sweet - having it only affect subsequent plugs... I mean, makes total sense in retrospect, but is perfect. Thanks for the quick README tweak, I'll just use that instead!

Issue #125, opened 2014-11-29
Handle Remote Plugins for neovim.


brendanjerwin:
http://neovim.org/doc/user/remote_plugin.html#remote-plugin

The Python sample works. I'd love to easily install the soon-to-be-a-torrent of awesome remote plugins for nvim.

As far as I can see, it'd need to know that its a remote plugin, place it into the right directory, and call :UpdateRemotePlugins after making changes.

Perhaps the Plugfile could come back to provide some of the needed meta-data about the plugin? (isRemotePlugin: true, host: python, etc...) If not that, then perhaps a `RemotePlug` function in addition to `Plug` would do the trick. (or, probably, both approaches would be best.)

fwalch:
One example of a remote plugin is [Floobits](https://floobits.com/help/plugins/nvim).

bruno-:
Hi,
I was just trying to setup neovim-floobits plugin locally with neovim and was totally confused to find out that none of the plugin commands (`:Floo...`) work..
So it seems the plugin is not sourced properly..

I found this issue via another one opened on neovim-floobits. I understand this is an "edgy" neovim feature that might not be supported immediately, but just wanted to let you know there are real people out there trying it out.

junegunn:
@bruno- I see, but the installation instruction of the plugin clearly states that the command is required. At this stage, I don't see it as any different from the other plugins that require extra steps in installation such as YCM or vim-proc. I'm not following the development of Neovim, so I don't know what the technical hurdles are, but requiring users to manually execute the command does not seem like the most elegant way possible and probably there's room for improvement.

(Sorry I posted this comment on the wrong issue)

starcraftman:
@junegunn I agree with your assessment. I'd just add that instructions also say a restart is required, I assume only for first install? So even if we added a built in hook, user would still be disrupted by restarting.

@bruno- If you just want automation, like YCM you can just use a simple do hook. Below works.

In your `.nvimrc`: 
`Plug 'Floobits/floobits-neovim', { 'do': function('hooks#floo') }`

Put this in ~/.nvim/autoload/hooks.vim
```vim
function! hooks#remote(info)
  if a:info.status == 'installed' || a:info.status == 'updated' || a:info.force
    UpdateRemotePlugins
    echom "Remember to restart!"
  endif
endfunction
```

Maybe we could put this into the troubleshooting faq?

bruno-:
Hi guys,
thank you for the quick response and helpful responses.

[...]

PR #450, opened 2016-03-21
support do vim command

lvht:
For post-update hooks, If you need to execute Vim's command, you can pass the command name with a leading `:`.
```
Plug 'Shougo/deoplete.nvim', {'do':':UpdateRemotePlugins'}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,6 +198,13 @@ Plug 'Shougo/vimproc.vim', { 'do': 'make' }
 Plug 'Valloric/YouCompleteMe', { 'do': './install.py' }
 ```
 
+If you need to execute Vim's command, you can pass the command name with a
+leading `:`.
+
+```
+Plug 'Shougo/deoplete.nvim', {'do':':UpdateRemotePlugins'}
+```
+
 If you need more control, you can pass a reference to a Vim function that
 takes a single argument.
 
--- a/plug.vim
+++ b/plug.vim
@@ -723,10 +723,15 @@ function! s:do(pull, force, todo)
       let type = type(spec.do)
       if type == s:TYPE.string
         try
-          " FIXME: Escaping is incomplete. We could use shellescape with eval,
-          "        but it won't work on Windows.
-          let g:_plug_do = '!'.escape(spec.do, '#!%')
-          execute "normal! :execute g:_plug_do\<cr>\<cr>"
+          if spec.do =~? '^:'
+            let g:_plug_do = spec.do[1:]
+            execute g:_plug_do
+          else
+            " FIXME: Escaping is incomplete. We could use shellescape with eval,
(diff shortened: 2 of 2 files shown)


junegunn:
Thanks. I guess you know it's already possible with funcref, right?

```vim
function! s:do()
  UpdateRemotePlugins
endfunction

Plug 'foo/bar', { 'do': function('s:do') }
Plug 'foo/baz', { 'do': function('s:do') }
```

It's a few more lines, but personally I prefer it this way as it's less confusing and more flexible.

Having said that, I think it's time we have to talk about `UpdateRemotePlugins` command itself (#125). I didn't like it in the first place, and I still don't think it's the most elegant, but we now seem to have some "remote" plugins that are actually used by people, and maybe it's time for vim-plug to consider doing it automatically.

junegunn:
It seems like Neovim team is [reconsidering `UpdateRemotePlugins`](https://github.com/neovim/neovim/pull/4438#issuecomment-195421685).
I'd like to wait until things are sorted out.

starcraftman:
@junegunn I don't know, they seem to lack a clear plan on plugins in general especially after they nixed/froze the discussion of their plugin manager/repository issue. I don't think there's any point doing much on the UpdateRemote command until they take some more concrete steps. I wish there was a feed for just the plugin/plugin manager changes don't regularly feel like trawling through neovim's issues.

Re this PR, I have no strong opinion.

junegunn:
@starcraftman True, I agree. I don't know where it's heading but one thing I can be sure is that they are looking for a better way.

Regarding this PR, it looks mostly okay but I have some concerns. Except for `:UpdateRemotePlugins`, have you seen any plugin that tells the user to execute some Vim command after update? If neovim guys decided to ditch `:UpdateRemotePlugins`, we are left with a dead feature. And since the future of the command is unclear, I'd rather not put it on the README page.

eduncan911:
@junegunn Yes, there are several plugins that ask to execute commands after first load.

`vim-go` is one:

> Please be sure all necessary binaries are installed (such as gocode, godef, goimports, etc.). You can easily install them with the included :GoInstallBinaries command. If invoked, all necessary binaries will be automatically downloaded and installed to your $GOBIN environment (if not set it will use $GOPATH/bin). Note that this command requires git for fetching the individual Go packages. Additionally, use :GoUpdateBinaries to update the installed binaries.

Though, deciding which one to run `GoInstallBinaries` vs `GoUpdateBinaries` during the vim-plug's `do` command is another topic of dicussion (most likely using a custom `funcref` is best to detect if already installed previously, then run `GoUpdateBinaries`).

junegunn:
Fixed in 17996ce. The committed code ensures that the commands from the plugin are available by loading the plugin beforehand.

PR #451, opened 2016-03-25
Reset colors for git show

simonweil:
Problem: When the user has non default color settings for git show,
         the output is full of terminal escaping sequences
Solution: Add `--no-color` to the `git show` command


junegunn:
Thanks, merged!

simonweil:
Thank you!

PR #463, opened 2016-04-03
Make PlugInstall!, PlugUpdate! trigger post-fetch checkout and submodule update

yous:
Using YouCompleteMe, I faced some cases that `PlugInstall` clones YouCompleteMe successfully, but submodules are not cloned by interrupt or some other failures. Then even if I do `PlugInstall!` or `PlugUpdate!`, vim-plug doesn't clone submodules, so I had to `cd ~/.vim/plugged/YouCompleteMe && git submodule update --init --recursive` either `rm -rf ~/.vim/plugged/YouComplete` followed by `PlugInstall` again.

This pull request makes `PlugInstall!` and `PlugUpdate!` clone submodules, but I'm not sure it's okay to make `filter` to include every item with `s:update.force`.

### DIFF
--- a/test/workflow.vader
+++ b/test/workflow.vader
@@ -721,6 +721,9 @@ Execute (On install):
   call plug#begin()
   Plug 'junegunn/vim-easy-align', { 'do': 'touch installed' }
   Plug 'junegunn/vim-pseudocl'
+  Plug 'junegunn/seoul256.vim'
+  Plug 'junegunn/goyo.vim'
+  Plug 'yous/subsubmodule'
   call plug#end()
 
   silent PlugInstall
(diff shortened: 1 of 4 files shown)


junegunn:
Thanks.

> but I'm not sure it's okay to make filter to include every item with s:update.force.

That actually makes the patch bigger than its description. We're about to extend the bang commands to force-trigger not only `git submodule update` but also `git checkout` and `git merge`. For one thing, the output of `:PlugDiff` will not be the same after `:PlugInstall!`. So, is it desirable? It's something to think about. It can be useful when you want to quickly switch branches or tags without fetching the remote.

1. Add `{ 'branch': 'something' }` to `Plug` command
2. Reload vimrc and `:PlugInstall!`

But are there cases where you *don't* want the behavior? (`:PlugInstall!` not retaining `:PlugDiff` output can be one thing, but I don't think it's something one can't live with especially when we explain the consequences of bang commands.)

Anyway we need to change the commit message if we decided to merge this.

yous:
Thanks for the clarification. I agree with you as I think doing `git checkout` to different branches or tags makes sense for `:PlugInstall!`. Then we should add additional tests and rename this pull request along with the commit message.

yous:
Updated with additional tests for changing branch, tag, or commit. If there are some documents that we have to update, please point them out.

junegunn:
:+1: 

I'm not completely sure what you mean by "work out of box" though. How about if we make the message more specific, something like "Make Plug[Install|Update]! trigger post-fetch checkout and submodule update". What do you think?

test/workflow.vader
```diff
@@ -752,44 +760,68 @@ Execute (On update):
 Execute (When already installed):
   call plug#begin()
   Plug 'junegunn/vim-easy-align', { 'do': 'touch installed2' }
-  Plug 'junegunn/vim-pseudocl', { 'do': 'touch installed2' }
```
junegunn: FYI, I specified two plugins here to make vim-plug start parallel installer. Now we have more plugins, we can safely remove it, which you did, nice.

yous:
Actually I was thinking about the commit message, struggling with 50 characters rule. It seems that GitHub wraps a commit message longer than 72 characters. Updated.

test/workflow.vader
yous: Oh I see. Actually didn't think about the parallel installation. Thanks for pointing it out.

junegunn:
Merged, thanks!

yous:
Thank you!

PR #467, opened 2016-04-12
Use python2 for YCM post-installation

hobarrera:
YouCompleteMe's post-installation script only seems to work with python2 (this is actually mentioned in their docs).

It seems that the sample documentation uses the system default python (which is no longer always python2), so let's better be explicit about it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -195,7 +195,7 @@ In that case, use `do` option to describe the task to be performed.
 
 ```vim
 Plug 'Shougo/vimproc.vim', { 'do': 'make' }
-Plug 'Valloric/YouCompleteMe', { 'do': './install.py' }
+Plug 'Valloric/YouCompleteMe', { 'do': 'python2 install.py' }
 ```
 
 If you need more control, you can pass a reference to a Vim function that
@@ -208,7 +208,7 @@ function! BuildYCM(info)
   " - status: 'installed', 'updated', or 'unchanged'
   " - force:  set on PlugInstall! or PlugUpdate!
   if a:info.status == 'installed' || a:info.force
-    !./install.py
+    !python2 install.py
   endif
 endfunction
 


junegunn:
Thanks!

vheon:
> YouCompleteMe's post-installation script only seems to work with python2 (this is actually mentioned in their docs).

This is no longer true since last month. If you have found docs stating differently let us know!

junegunn:
@vheon Should we revert the change then?

hobarrera:
That's odd. It didn't work for me when building with python3 (YCM simply "did not work"). I tried this ~12hours ago.

vheon:
> "did not work"

And by that I'm assuming you mean that you've build YCM but then it wasn't working correctly?

hobarrera:
Yes; it built, by no auto-completions were ever suggested. I saw no errors anywhere either.

At the time, I recalled having read about the py2 vs py3 thing in YCM's doc, retried building with each a few times, and attributed it. How can I provide further details on that?

vheon:
> Yes; it built, by no auto-completions were ever suggested. I saw no errors anywhere either.

So you've enabled the logs and there is anything? that is really strange but be careful because when you build YCM with Python3 but try to run it from a vim which has only Python2 support then ycmd (the server) will crash as soon as it starts. Look at the docs for this kind of scenarios https://github.com/Valloric/YouCompleteMe#i-get-importerror-exceptions-that-mention-pyinit_ycm_core-or-initycm_core

hobarrera:
Odd, logs reveal that YCM is attempting to load with python2:

    ImportError: dynamic module does not define init function (initycm_core)

But I haven't configured it to use `python2`. Is there any reason why it would use the non-default python interpreter? It feels odd to have to configure it to use the default python (`/usr/bin/python`, which is `python3`).

vheon:
As I've said earlier you have to look at the vim's version and check which version is compiled against. From the error I guess that vim is compiled with the Python2 interpreter, so the YCM client run under Python2 and when spawning the server it uses the same interpreter, unless told otherwise with the appropriate option, but now you're starting the the ycmd server with Python2 while the compiled part of the server is compiled with Python3 (the thing is not much a different version but a different interpreter, so even if you compiled the server with the same version of Python2 that vim uses but not the same executable it would crash), so when ycmd tries to load the compiled libraries it crashes.

> @vheon Should we revert the change then?

@junegunn yes

junegunn:
@vheon Okay, thanks. I know you are one of the maintainers of YCM so I should believe you :)
@hobarrera Reverted the change, but thanks anyway.

hobarrera:
@vheon Nope, vim seems to be built with python3.:
```
$ vim --version | grep py
+cryptv          +linebreak       +python/dyn      +vreplace
+cscope          +lispindent      +python3/dyn     +wildignore
```

`python3` is the system default python, there are no references to python2, nor in my config, nor anywhere. [vim is built with python3](https://projects.archlinux.org/svntogit/packages.git/tree/trunk/PKGBUILD?h=packages/vim) too.

(for the record, I don't have any weird builds or anything, this is all clean packages from my distro too)

---

@junegunn
Yup, that makes sense. If ycm continues to fail for me, I'll try to follow this up on the ycm-side, but keeping the example code here as-is seems logical. Thanks.

vheon:
> @vheon Nope, vim seems to be built with python3.:
>
$ vim --version | grep py
+cryptv          +linebreak       +python/dyn      +vreplace
+cscope          +lispindent      +python3/dyn     +wildignore
python3 is the system default python, there are no references to python2, nor in my config, nor anywhere. vim is built with python3 too.

Actually since you have both `+python/dyn` `+python3/dyn` and you're on Archlinux it can use both at the same time, and since I don't want to go into that rabbit hole I suggest you just use the ycmd setting to point to the right python which compiled ycmd and call it the day 😜

PR #476, opened 2016-04-17
Support wildcards in tag option

### LINKED ISSUE #427: Add option to track latest tagged release
The `tag` Plug option is very nice for being able to use the latest stable release of a plugin. However, it also essentially removes the autoupdate functionality of vim-plug. This is fine if you're using it to stick with a known good version of the plugin, but less good if you're just trying to stay on the latest stable version. Many vim plugins do not have a separate development branch that can be tracked.

Therefore, I suggest the addition of a `track` Plug option, which can be set to always check out the most recent tag, as determined by `git tag --list --sort -version:refname | head -1` (or similar).

The syntax of this command is obviously up to you, perhaps there is a nice way it could be worked in to specify either this or the current behavior (e.g. `'track': 'branch'`, `'track': 'tags'`, etc). This may or may not be a useful way to express this. Alternatively, perhaps some sort of "magic" value of the existing `tag` option could be used, though this seems trickier from a collision standpoint.

Thanks for your consideration!

```
:version
NVIM 0.1.2
Build type: RelWithDebInfo
Compilation: /usr/local/Library/ENV/4.3/clang -Wconversion -O2 -g -DDISABLE_LOG -Wall -Wextra -pedan
tic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wvla -fstack-protector-strong -fdiagnostic
s-color=auto -DINCLUDE_GENERATED_DECLARATIONS -DHAVE_CONFIG_H -I/tmp/neovim20160211-20191-1u5l9yc/ne
ovim-0.1.2/build/config -I/tmp/neovim20160211-20191-1u5l9yc/neovim-0.1.2/src -I/tmp/neovim20160211-2
0191-1u5l9yc/neovim-0.1.2/deps-build/usr/include -I/tmp/neovim20160211-20191-1u5l9yc/neovim-0.1.2/de
ps-build/usr/include -I/tmp/neovim20160211-20191-1u5l9yc/neovim-0.1.2/deps-build/usr/include/luajit-
2.0 -I/tmp/neovim20160211-20191-1u5l9yc/neovim-0.1.2/deps-build/usr/include -I/tmp/neovim20160211-20
191-1u5l9yc/neovim-0.1.2/deps-build/usr/include -I/tmp/neovim20160211-20191-1u5l9yc/neovim-0.1.2/dep
s-build/usr/include -I/tmp/neovim20160211-20191-1u5l9yc/neovim-0.1.2/deps-build/usr/include -I/usr/l
ocal/opt/gettext/include -I/usr/include -I/usr/include -I/tmp/neovim20160211-20191-1u5l9yc/neovim-0.
1.2/build/src/nvim/auto -I/tmp/neovim20160211-20191-1u5l9yc/neovim-0.1.2/build/include
Compiled by vil@featherweight

Optional features included (+) or not (-): +acl   +iconv    +jemalloc
For differences from Vim, see :help vim-differences

   system vimrc file: "$VIM/sysinit.vim"
  fall-back for $VIM: "/usr/local/Cellar/neovim/0.1.2/share/nvim"
```

- Type:
    - [ ] Bug
    - [ ] Enhancement
    - [x] Feature Request
    - [ ] Question
- OS:
    - [ ] All/Other
    - [ ] Linux
    - [x] OS X
    - [ ] Windows
- Vim:
    - [ ] Terminal Vim
    - [ ] GVim
    - [x] Neovim

junegunn:
Close #427

- No manual parsing of the expression; just uses `git tag --list <pattern>` for wildcard expansion
- *ProTip*: `{ 'tag: '*' }` will point to the latest tagged release
- Requires git 1.9.2 or above (`error: unknown option 'sort'`)
    - Are we fine with this? I think it's okay, as it's probably not going to be the most popular option.
    - So the test case fails on travis-ci as it comes with git 1.8.5.6.
         - We can either install the latest git using ppa, or just skip the test if the version requirement is not met
         - Since I'd like to make sure that vim-plug is still functional with older versions of git and Vim, so let's go with the latter.

/cc @vilhalmer

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,6 +89,9 @@ Plug 'tpope/vim-fireplace', { 'for': 'clojure' }
 " Using a non-master branch
 Plug 'rdnetto/YCM-Generator', { 'branch': 'stable' }
 
+" Using a tagged release; wildcard allowed (requires git 1.9.2 or above)
+Plug 'fatih/vim-go', { 'tag': '*' }
+
 " Plugin options
 Plug 'nsf/gocode', { 'tag': 'v.20150303', 'rtp': 'vim' }
 
--- a/plug.vim
+++ b/plug.vim
@@ -28,6 +28,9 @@
 "   " Using a non-master branch
 "   Plug 'rdnetto/YCM-Generator', { 'branch': 'stable' }
 "
+"   " Using a tagged release; wildcard allowed (requires git 1.9.2 or above)
+"   Plug 'fatih/vim-go', { 'tag': '*' }
+"
 "   " Plugin options
 "   Plug 'nsf/gocode', { 'tag': 'v.20150303', 'rtp': 'vim' }
 "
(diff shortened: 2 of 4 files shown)


vilhalmer:
Excellent, thanks so much!

PR #497, opened 2016-06-26
Fix install plugins in Cygwin or MSYS2 vim with enabled git core.auto…

rikdev:
Vim from Cygwin or MSYS2 require LF line endings in plugins. But option core.autocrlf for Git for Windows usually is true and git converts LF to CRLF. This PR disabled core.autocrlf option for git.

### DIFF
--- a/plug.vim
+++ b/plug.vim
@@ -931,7 +931,7 @@ function! s:update_impl(pull, force, args) abort
   silent! redraw
 
   let s:clone_opt = get(g:, 'plug_shallow', 1) ?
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks for the patch. Unfortunately I don't have a Windows machine and I can't verify the validity of the approach. Any comments from Windows users are very welcome.

Having said that, I don't think the patch is logically correct. It does not append the flag when `g:plug_shallow` is 0, which is unrelated to the issue.

PR #508, opened 2016-07-31
Silence :delete commands

mhinz:
Add `:silence` to all `:delete` commands to avoid filling up the message history with *"1 line less"* messages.

---

I noticed this after `:PlugUpdate`, so silencing the `:delete` in `s:log()` would have been enough. But I found 2 other unsilenced occurences and couldn't come up with a reason why they shouldn't be silenced as well.

### DIFF
--- a/plug.vim
+++ b/plug.vim
@@ -1048,14 +1048,14 @@ function! s:update_finish()
       if v:shell_error
         call add(s:update.errors, name)
         call s:regress_bar()
-        execute pos 'd _'
+        silent execute pos 'd _'
         call append(4, msg) | 4
       elseif !empty(out)
         call setline(pos, msg)
       endif
       redraw
     endfor
-    4 d _
+    silent 4 d _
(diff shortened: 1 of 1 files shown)


junegunn:
Any idea why I don't see the message? I get no message on `:d _` or `:d _ 2`. I start to see `3 fewer lines` once the count exceeds 2.

mhinz:
For me the behaviour is like this:

- **empty buffer**: `<no output>`
- **1 line in buffer**: `--No lines in buffer--`
- **2+ lines in buffer**: `1 line less`

UPDATE: Oha, it's because of `'report'`. Totally forgot about that option.

junegunn:
Ah, okay, I didn't know about the option. How about `silent execute` instead of `execute 'silent'`? We save two bytes per line :)

mhinz:
> We save two bytes per line :)

Done. ;)

I also mentioned `'report'` in the long message, to be on the safe side.

junegunn:
Thank you!

PR #522, opened 2016-09-15
Fixing E684 when VisualHostKey is enabled for ssh

jmatth:
Having VisualHostKey enabled for ssh can cause lines containing a single "^M" character to appear in the git output, which causes E684 to be thrown if not filtered out.

This just changes a filter to ignore any sort of newline characters so that the following split should never return an empty list.

### DIFF
--- a/plug.vim
+++ b/plug.vim
@@ -1089,7 +1089,7 @@ function! s:job_handler(job_id, data, event) abort
 
   if a:event == 'stdout'
     let complete = empty(a:data[-1])
-    let lines = map(filter(a:data, 'len(v:val) > 0'), 'split(v:val, "[\r\n]")[-1]')
+    let lines = map(filter(a:data, 'v:val =~ "[^\r\n]"'), 'split(v:val, "[\r\n]")[-1]')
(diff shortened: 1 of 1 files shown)


plug.vim
```diff
@@ -1089,7 +1089,7 @@ function! s:job_handler(job_id, data, event) abort
 
   if a:event == 'stdout'
     let complete = empty(a:data[-1])
-    let lines = map(filter(a:data, 'len(v:val) > 0'), 'split(v:val, "[\r\n]")[-1]')
+    let lines = map(filter(a:data, 'len(substitute(v:val, "[\r\n]", "", "g")) > 0'), 'split(v:val, "[\r\n]")[-1]')
```
junegunn: How about using `v:val =~ "[^\r\n]"` instead?

jmatth: That works too, and even handles filtering the empty string so `len` isn't needed anymore.

jmatth:
Made the change from `substitute` to `=~ "[^\r\n]"`

junegunn:
Merged to master. Thanks!

PR #527, opened 2016-09-27
Add option to disable automatic submodule pulling

pnloyd:
Add an option "submod" that's allows disabling cloning of sub-modules. Default behavior is the same as before.

I have run into two use cases.
1. Some plugins (e.g YCM) have many sub-modules required only for optional functionality. It would be nice to not have to download these on updates.
2. Some projects (e.g Omnisharp-vim) reference old versions of the sub-modules. This option would allow the user to still use vim-plug's update functionality and manually handle issues with sub-module references.

### DIFF
--- a/plug.vim
+++ b/plug.vim
@@ -57,6 +57,7 @@
 "| `on`                    | On-demand loading: Commands or `<Plug>`-mappings |
 "| `for`                   | On-demand loading: File types                    |
 "| `frozen`                | Do not update unless explicitly specified        |
+"| `submod`                | Recursively fetch sub-repositories (submodules)  |              |
 "
 " More information: https://github.com/junegunn/vim-plug
 "
@@ -99,7 +100,7 @@ let s:mac_gui = has('gui_macvim') && has('gui_running')
 let s:is_win = has('win32') || has('win64')
 let s:nvim = has('nvim') && exists('*jobwait') && !s:is_win
 let s:me = resolve(expand('<sfile>:p'))
-let s:base_spec = { 'branch': 'master', 'frozen': 0 }
+let s:base_spec = { 'branch': 'master', 'frozen': 0, 'submod': 1 }
 let s:TYPE = {
 \   'string':  type(''),
 \   'list':    type([]),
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks for the contribution. I'd like to hear what others think about the addition. It's a small change in the code, indeed, but what I'm not particularly happy about is that it may matter only to a very limited number of plugins. You presented two use cases, but the latter sounds like it's a problem that the plugin itself has to fix, rather than something that a plugin manager should address. Then we are left with YCM, the singularity, that has always been demanding more options and special care. I can't think of another plugin right now. So I'm a bit torn on this one.

Anyway regarding the pull request:

- Extraneous `|` in the table. The description should also be on the README file.
- `submod` is only accessed via `get()` with a default value, so we don't need it in `s:upgrade_specs()`
- A test case would be nice. You probably have no experience with vader test framework, I can help you with this.

@vheon Can you share your thoughts on this, as one of the maintainers of YCM? If we skip the submodule update, can YCM detect that and do it later by itself?

vheon:
> Some plugins (e.g YCM) have many sub-modules required only for optional functionality. It would be nice to not have to download these on updates.

Here is the thing: YCM doesn't have many sub-modules it just have 2 (https://github.com/Valloric/YouCompleteMe/blob/master/.gitmodules) all the submodule that I think @pnloyd is talking about are submodule of `ycmd`. So the actual solution to your problem isn't to remove the `git submodule update --init --recursive` but rather to remove the `--recursive` flag. Now the question would be: does this impact other plugin other than YCM? And is this worth adding to `vim-plug` itself? Wouldn't be easier to just manage the plugin yourself as a local plugin?

> @vheon Can you share your thoughts on this, as one of the maintainers of YCM? If we skip the submodule update, can YCM detect that and do it later by itself?

`ycmd` checks if the things it needs are being pulled and suggests to run `git submodule update --init --recursive` when it thinks is necessary but it doesn't do it for you.

junegunn:
@vheon Thanks for the reply.

> Wouldn't be easier to just manage the plugin yourself as a local plugin?

Hmm, yeah that actually sounds like a good strategy we can suggest. I added a section to the wiki page.

https://github.com/junegunn/vim-plug/wiki/faq#installing-youcompleteme-manually

pnloyd:
I can't think of amy other plugins that have this issue.. It seems manual
management of the plugin would be a good solution for both cases.

blueyed:
Sounds like `submod` from this PR could be 3-state: `no submodules`, `non-recursive submodules` (maybe the new default) and then `recursive submodules`?!

vheon:
> non-recursive submodules

To me this doesn't make sense, I mean a plugin manager should either fully manage a plugin or let the user handle everything (that is why I suggested to manage the plugin as a local plugin).

pnloyd:
Closing this since the common consensus seems to be that the option is not especially useful.

PR #529, opened 2016-10-05
Add environment variable for use with 'do':

jbro:
I've found it handy to be able to reference the path set in plug#begin from the 'do' command. So I've introduced the environment variable $VIMPLUG_DIR.

### DIFF
--- a/plug.vim
+++ b/plug.vim
@@ -126,6 +126,8 @@ function! plug#begin(...)
   let g:plugs_order = []
   let s:triggers = {}
 
+  let $VIMPLUG_DIR = home
+
   call s:define_commands()
   return 1
 endfunction


vheon:
Can you make some example? Why can't you use the `g:plug_home`?

jbro:
I use it like

    Plug 'vim-scripts/YankRing.vim', { 'do': 'mkdir -p $VIMPLUG_DIR/../tmp' }

and

    Plug 'powerline/fonts', { 'dir': '$VIMPLUG_DIR/../fonts/powerline', 'do': './install.sh' } |
      Plug 'powerline/powerline', { 'tag': 'master', 'rtp': 'powerline/bindings/vim/' }

But I just realised that I can just do:

        Plug 'vim-scripts/YankRing.vim', { 'do': 'mkdir -p '+ g:plug_home +'/../tmp' }

It might still be useful for scripts to detect if they are run by vim-plug.

junegunn:
Hi, thanks for the suggestion but I don't think we have to add it to vim-plug.

vim-plug changes the working directory before executing post-update hook of each plugin so the first example you gave above can be rewritten as follows

```vim
Plug 'vim-scripts/YankRing.vim', { 'do': 'mkdir -p ../../tmp' }
```

And yes, we also have `g:plug_home`:

```vim
Plug 'powerline/fonts', { 'dir': g:plug_home.'/../fonts/powerline', 'do': './install.sh' }
```

If you don't like such string concatenation and if you do have a use case where some script has to know if it's run by vim-plug, then environment variable can be useful but it doesn't have to reside in plug.vim.

```vim
call plug#begin('~/.vim/plugged')
let $VIMPLUG_HOME = g:plug_home
```

The above code does the exact same thing, and it's more flexible. If we can do something without adding a new feature (and without too much extra effort), then we prefer do it that way and keep vim-plug small and simple.

jbro:
No problem, your suggestions are much cleaner solutions :-)

PR #532, opened 2016-10-08
Use :enew instead of :new when being the only window

blueyed:
After `:only` in the vim-plug window, `:PlugUpdate` currently results in
a split window, but I think it is better to keep it as the single window
in that case.

### DIFF
--- a/plug.vim
+++ b/plug.vim
@@ -726,10 +726,17 @@ function! s:prepare(...)
 
   call s:job_abort()
   if s:switch_in()
+    let new_window = winnr('$') > 1
     normal q
+  else
+    let new_window = 1
   endif
 
-  call s:new_window()
+  if new_window
+    call s:new_window()
+  else
+    enew
+  endif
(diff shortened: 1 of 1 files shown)


junegunn:
Makes sense, thanks. What do you think about the following diff?

```diff
diff --git a/plug.vim b/plug.vim
index 1da64a9..d4ab48a 100644
--- a/plug.vim
+++ b/plug.vim
@@ -726,10 +726,11 @@ function! s:prepare(...)
 
   call s:job_abort()
   if s:switch_in()
-    normal q
+    enew
+  else
+    call s:new_window()
   endif
 
-  call s:new_window()
   nnoremap <silent> <buffer> q  :if b:plug_preview==1<bar>pc<bar>endif<bar>bd<cr>
   if a:0 == 0
     call s:finish_bindings()
```

blueyed:
Your patch is better, thanks!
I'd say to commit it and close this PR through it.

junegunn:
I need to add a test case for it. I'll push it tomorrow, thanks.

PR #534, opened 2016-10-13
use job_start on vim

mattn:

### DIFF
--- a/plug.vim
+++ b/plug.vim
@@ -1163,9 +1163,9 @@ function! s:nvim_job_handler(job_id, data, event) abort
   endif
 
   if a:event == 'stdout'
-    call s:job_out_cb(self, a:job_id, data)
+    call s:job_out_cb(self, a:job_id, a:data)
   elseif a:event == 'exit'
-    call s:job_exit_cb(self, a:job_id, data)
+    call s:job_exit_cb(self, a:job_id, a:data)
   endif
 endfunction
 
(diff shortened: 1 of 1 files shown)


mattn:
@junegunn Sorry, I can't figure out what is wrong in last one error. Can you handle this?

plug.vim
```diff
@@ -1085,58 +1087,98 @@ function! s:update_finish()
...
-    let self.result = join(self.lines, "\n")
-    if !complete
-      call remove(self.lines, -1)
-    endif
-    " To reduce the number of buffer updates
-    let self.tick = get(self, 'tick', -1) + 1
-    if self.tick % len(s:jobs) == 0
-      call s:log(self.new ? '+' : '*', self.name, self.result)
-    endif
+    call s:job_out_cb(self, a:job_id, data)
```
mattn: right. thanks.

junegunn:
Cool, I'll take a look later in the day. Note that vader.vim, the test framework we use here, is currently unable to test asynchronous installer, so we can see that the travis builds are not running correctly except for the first one where Vim 7.2 is used (`ENV=nox`). I'll also check if I can extend vader to support async features.

junegunn:
@mattn Sorry for the delay. I decided that we need to set up proper automated testing for async installers. We already have Neovim installer that is not at all covered by the tests and I don't feel good about adding another one that is also not tested. It means that I have to extend vader.vim for asynchronous behaviors. I have some ideas and I'll let you know how it turns out.

mattn:
Sure. let me known if the time come.

junegunn:
Alright, I think I got it working. All the tests are now passing.
https://travis-ci.org/junegunn/vim-plug/builds/168217454

I made a few changes to your patch. Can you review the commits in https://github.com/junegunn/vim-plug/commits/vim8 ?

mattn:
Working good for me!

vim-plug with job on vim8

junegunn:
Thanks, good to see that it's working correctly on Windows as well. I'll review the changes again and merge it to master soon.

PR #545, opened 2016-10-26
Set nolist in vim-plug buffers

srstevenson:
In the buffer opened by `:PlugInstall`, lines streaming the output from `git clone` have trailing whitespace, which is highlighted by default in Neovim when `list` is enabled. By disabling `list` in the buffer, this whitespace is no longer highlighted.


junegunn:
Thanks.

Issue #480, opened 2016-04-28
vim-plug fails to PlugUpdate if no master branch exists


crowell:
Managing `Plug 'rdnetto/YCM-Generator'` with `:PlugUpdate` fails with this error

`x YCM-Generator: error: pathspec 'master' did not match any file(s) known to git.`

YCM-Generator only has https://github.com/rdnetto/YCM-Generator stable and develop branches, not master. Vim-Plug assumes that i want master, which isn't the case. Whatever the current in-use branch is, is the one I want.

---

I'm using the latest from master now

```
41de3c713c0b82700bfcb50a45e9d7756a2cce96
```

- Type:
  - [x] Bug
  - [ ] Enhancement
  - [ ] Feature Request
  - [ ] Question
- OS:
  - [x] All/Other
  - [ ] Linux
  - [ ] OS X
  - [ ] Windows
- Vim:
  - [x] Terminal Vim
  - [x] GVim
  - [x] Neovim

vheon:
This is a duplicate of #288 and quoting from https://github.com/rdnetto/YCM-Generator#installation

> For vim-plug users, add Plug 'rdnetto/YCM-Generator', { 'branch': 'stable'} to your vimrc.

junegunn:
The idea was to prevent the case where the user manually checks out a non-default, experimental branch, forgets about it, and vim-plug stops updating the plugin. By explicitly specifying the name of the branch with `master` as the default allows vim-plug to easily verify the status of the clone. The obvious downside of this approach is that it doesn't work well with plugins that do not use `master` as their default branches. But since there are very few such plugins, I didn't bother to find a proper solution to it. I wonder if there's a way to find the default branch without contacting the remote again and without external bookkeeping.

blueyed:
Using `@{upstream}` might help here (via #546).

john-kurkowski:
For GitHub at least, there is a gradual move away from the branch name `master` to `main`, described in [github/renaming](https://github.com/github/renaming). Just had to add `{ 'branch': 'main' }` to a `Plug` myself. Would it be easy to support that 1 exception?

junegunn:
The latest version of vim-plug tracks the default branch, so this should be fixed now.

PR #546, opened 2016-10-27
git_validate: better error with checkout being ahead/diverged

blueyed:
The `--count` and `--left-right` options seem to be in Git since before
1.7 already.

The error message is improved, and distinguishes between only being
ahead (which is the case if you could push), or being diverged (ahead
and defined).

I think it could be bad to accidentally clean/delete a local checkout
which is ahead, and therefore PlugClean is not mentioned then, which
appears to trigger it to show up in :PlugClean.

btw: what about using `@{upstream}` (Git 1.7.0) more / in general?

### DIFF
--- a/plug.vim
+++ b/plug.vim
@@ -2018,18 +2018,15 @@ function! s:git_validate(spec, check_branch)
               \ a:spec.branch), a:spec.dir)), '\t')
         if !v:shell_error && ahead
           if behind
-            let base_error = printf(
-                  \ 'Diverged from origin/%%s (%%d commit(s) ahead and %d commit(s) behind!', behind)
+            " Only mention PlugClean if diverged, otherwise it's likely to be
+            " pushable (and probably not that messed up).
+            let err = printf(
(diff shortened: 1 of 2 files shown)


junegunn:
Looks good. I'd like to simplify the messages a bit.

```
Diverged from origin/%s: %d commit(s) ahead and %d commit(s) behind.
Backup local changes and run PlugClean and PlugUpdate to reinstall it.
```

```
Ahead of origin/%s by %d commit(s).
Cannot update until local changes are pushed.
```

And can you try writing test cases?

> btw: what about using @{upstream} (Git 1.7.0) more / in general?

I didn't know about `@{upstream}`. What problem does it solve? I don't see a reason to change if it doesn't fix anything. Can it be a solution to #480?

blueyed:
Done, pushed a fixup.

> And can you try writing test cases?

Of course.. :)

junegunn:
Merged, thanks for the contribution!

PR #547, opened 2016-10-27
Display not-loaded status with all plugins that have a dir

blueyed:
I think that is useful information, especially given that the error only comes from `s:git_validate` and is about different branches etc in most cases(?).


blueyed:
@junegunn 
Anything holding this back?

junegunn:
Sorry for not responding to this earlier. It just slipped my mind, probably because I couldn't see how useful the additional information would be to the users, but the change is small and it looks good to me.

PR #548, opened 2016-10-29
Deferred loading on autocmd (#event[#pattern])

junegunn:
Does not mean that I'm positive on this. I'm just posting the patch as it was trivial to implement the idea.

```vim
Plug 'SirVer/ultisnips',  { 'on': '#InsertEnter' }
Plug 'junegunn/goyo.vim', { 'on': ['Goyo', '#BufEnter#README,BUILD,CHANGELOG'] }
```

(See `:help exists()` for `#event#pattern` syntax)

### DIFF
--- a/plug.vim
+++ b/plug.vim
@@ -229,9 +229,16 @@ function! plug#end()
             call s:assoc(lod.cmd, cmd, name)
           endif
           call add(s:triggers[name].cmd, cmd)
+        elseif cmd[0] == '#' && exists('##'.split(cmd, '#')[0])
+          let tokens = split(cmd, '#')
+          let group = 'Plug/'.name
+          execute 'augroup' group
+          autocmd!
+          execute 'autocmd' tokens[0] get(tokens, 1, '*') printf('call s:lod_autocmd(%s)', string(name))
+          execute 'augroup END'
         else
           call s:err('Invalid `on` option: '.cmd.
-          \ '. Should start with an uppercase letter or `<Plug>`.')
+          \ '. Should start with an uppercase letter, `<Plug>`, or `#`.')
         endif
       endfor
     endif
@@ -443,9 +450,7 @@ function! plug#load(...)
     let s = len(unknowns) > 1 ? 's' : ''
     return s:err(printf('Unknown plugin%s: %s', s, join(unknowns, ', ')))
   end
-  for name in a:000
-    call s:lod([name], ['ftdetect', 'after/ftdetect', 'plugin', 'after/plugin'])
-  endfor
+  call s:lod(a:000, ['ftdetect', 'after/ftdetect', 'plugin', 'after/plugin'])
   call s:dobufread(a:000)
   return 1
 endfunction
@@ -524,6 +529,14 @@ function! s:lod_map(map, names, with_prefix, prefix)
     call feedkeys(prefix, 'n')
   endif
   call feedkeys(substitute(a:map, '^<Plug>', "\<Plug>", '') . extra)
+endfunction
+
+function! s:lod_autocmd(name)
+  call s:lod([a:name], ['ftdetect', 'after/ftdetect', 'plugin', 'after/plugin'])
+  call s:dobufread([a:name])
+  let group = 'Plug/'.a:name
+  execute 'autocmd!' group
+  execute 'augroup!' group
 endfunction
 
 function! plug#(repo, ...)
--- a/README.md
+++ b/README.md
@@ -121,16 +121,16 @@ Reload .vimrc and `:PlugInstall` to install plugins.
 
 ### `Plug` options
 
-| Option                  | Description                                      |
-| ----------------------- | ------------------------------------------------ |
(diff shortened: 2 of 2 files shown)


vheon:
Since which version is the `#event#pattern` syntax supported? Just asking because you don't seems to have a minimum version of vim for vim-plug and I wouldn't want you to break some users :)

junegunn:
It has been quite some time.

https://github.com/vim/vim/commit/f4cd3e8074641af68bf2b6a8579c3da58f0ac013
https://github.com/vim/vim/commit/a9b1e74b5d1d59c0a071ccacfe9aaaca9b4a5cca

The code in this PR only uses `exists('##EventName')`. `#event#pattern` syntax was chosen not to introduce a new, arbitrary syntax, but it's not directly used. The code splits the string.

vheon:
Thanks for digging that up! ;) As usual I hope that there is a real use for this other than YCM (I know you get a lots of requests from it).

plug.vim
```diff
@@ -526,6 +531,14 @@ function! s:lod_map(map, names, with_prefix, prefix)
   call feedkeys(substitute(a:map, '^<Plug>', "\<Plug>", '') . extra)
 endfunction
 
+function! s:lod_autocmd(name)
+  call s:lod([a:name], ['ftdetect', 'after/ftdetect', 'plugin', 'after/plugin'])
+  call s:dobufread([a:name])
```
blueyed: This might be problematic.
I found this PR when looking for a better way to handle custom `InsertEnter` handling.

I am using the following:
```vim
    Plug 'Shougo/deoplete.nvim', {'on': []}
    …
    augroup vimplug_load_on_insertmode
      autocmd!
      autocmd InsertEnter * call plug#load(
            \ 'cursorcross.vim',
            \ 'delimitMate',
            \ 'deoplete.nvim',
            \ ) | call deoplete#enable()
            \ | autocmd! vimplug_load_on_insertmode
    augroup END
```

And here the call to `deoplete#enable()` adds a `BufEnter` autocmd that gets triggered then.
It is a bit weird, since `plug#load()` should be finished already, before `deoplete#enable()` then sets up the `BufEnter` autocmd?!  (I have also put it into a separate function to make sure `|` does not cause it).

Maybe the `BufEnter` gets queued after the `InsertEnter`?!

blueyed: The `dobufread` happens for `ftplugin …/.vim/plugged/jedi-vim//**`, which is triggered through another handler:
```vim
    function s:my_insertenter_python()
      if &ft ==# 'python'
        call plug#load(
            \ 'deoplete-jedi',
            \ 'jedi-vim',
            \ )
        autocmd! vimplug_load_on_insertmode_python
      endif
    endfunction
    augroup vimplug_load_on_insertmode_python
      autocmd!
      autocmd InsertEnter * call s:my_insertenter_python()
    augroup END
```

blueyed: So, this explains the issue: there are two calls to `plug#load` in that setup.

blueyed:
I think this would be good to have.

But for my use case I would need to be able to a) call a custom function afterwards (`deoplete#enable()`), and b) would like to limit it to a filetypes also (i.e. `#InsertEnter`, but only for `python` filetypes).

blueyed:
> call a custom function afterwards

That is also useful for lazy-loading Unite, so that you can customize it using `call unite#custom#source(…)` etc.

junegunn:
@blueyed 

> call a custom function afterwards 

vim-plug currently supports that via User autocmd

```vim
" Code to execute when the plugin is lazily loaded on demand
Plug 'junegunn/goyo.vim', { 'for': 'markdown' }
autocmd! User goyo.vim echom 'Goyo is now loaded!'
```

Have you tried it?

> b) would like to limit it to a filetypes also (i.e. #InsertEnter, but only for python filetypes).

Hmm, I certainly wouldn't want to go that far. In my opinion, custom autocmd is the answer to such complex use cases.

blueyed:
@junegunn 
> vim-plug currently supports that via User autocmd

Oh, I missed that, sorry.  That is nice.
But I wonder if the autocommand name should not have something like a `plug_loaded_` prefix?!
Otherwise it could conflict/overlap with a plugin's own custom `User` autocommand?!

Thanks!

jbro:
I think this is a neat little feature, I'm currently using:

    Plug 'vim-syntastic/syntastic', { 'on': [] }
    "Delay syntatic load until we aren't doing anything
    augroup LazySyntatic
      autocmd!
      autocmd CursorHold * :call plug#load('syntastic')
      autocmd CursorHold * :autocmd! LazySyntatic
    augroup END

to delay the loading of syntastic. With  deffered autoload that could be hidden away with:

    Plug 'vim-syntastic/syntastic', { 'on': '#CursorHold' }

SwenChan:
So, can plug support event now? I'm using SpaceVim which using 'dein' as their default plugin manager and I want to switch to plug , which is my favorite vim plugin manager. But I need this feature because so many plugin in SpaceVim need this.
So, anyone can help with this feature? @junegunn @blueyed

akkarapon26 (approved):

PR #558, opened 2016-12-03
Do not check spelling in vim-plug buffers

HiPhish:
I added `nospell` to the settings of the `vim-plug` buffer. Since the user never types anything into this buffer and plugins can have weird names along with professional jargone there is no point in spell-checking anything.


junegunn:
Merged, thanks.

Travis failure is unrelated to the change. https://github.com/travis-ci/travis-ci/issues/6963

PR #562, opened 2016-12-10
test/run: shellcheck fixes

blueyed:


junegunn:
Thanks, how about if we just `set -eu`? So that things like `|| exit` are not needed.

blueyed:
Amended.

test/run
```diff
@@ -39,13 +43,13 @@ clone_repos() (
 )
 
 make_dirs() (
-  rm -rf "$PLUG_FIXTURES/$1"
+  rm -rf "${PLUG_FIXTURES:?}/$1"
```
junegunn: I believe this is also covered by `set -u`.

junegunn:
Sorry about the delay. Thanks!

PR #566, opened 2016-12-13
Neovim: use "dict" when referring to "self"

mhinz:
Recent Neovim added a huge PR that made its Partial support on par with Vim's.

Unexpected side-effect: Neovim was less strict than Vim about the lack of `dict` if the function referred to `self`.

Please see: https://github.com/neovim/neovim/issues/5763

### DIFF
--- a/plug.vim
+++ b/plug.vim
@@ -1153,7 +1153,7 @@ function! s:job_cb(fn, job, ch, data)
   call call(a:fn, [a:job, a:data])
 endfunction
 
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks, I'll merge when the tests pass. By the way, is https://github.com/junegunn/fzf/issues/765 related?

mhinz:
> By the way, is junegunn/fzf#765 related?

Argh, yes, same problem.

Issue #586, opened 2017-01-30
Windows cmd bug when executing utf8 .bat file


slepogin:
Hi,
I have this settings in _vimrc
lan mes ru_RU.UTF-8
set termencoding=utf8
set encoding=utf8
set fileencoding=utf8

I have utf8 symbols in vim-plag plugins directory path.
So when vim-plug creates .bat file for job_start it saves it in utf-8 and path in this file displayes correctly.

bat content:
git clone --depth 1 --no-single-branch  "https://git::@github.com/vim-airline/vim-airline.git" "C:\Users\Никита\.vim\plugged\vim-airline" 2>&1

But ofc windows cmd uses cp866 by default and when I trying to reproduce vim-plug's behaive using this command:
cmd /c C:\Users\Никита\AppData\Local\Temp\VIA383D.tmp.bat

I can see the issue same as in vim-plug:
C:\Users\Никита>git clone --depth 1 --no-single-branch  "https://git::@github.com/vim-airline/vim-airline.git" "C:\Users
\╨Э╨╕╨║╨╕╤В╨░\.vim\plugged\vim-airline"  2>&1
fatal: could not create leading directories of 'C:\Users\╨Э╨╕╨║╨╕╤В╨░\.vim\plugged\vim-airline': Permission denied

The only workaround I have found so far is to put
chcp 65001
in batch or in command line.

------------------------------

```
VIM - Vi IMproved 8.0 (2016 Sep 12, compiled Nov  6 2016 17:20:44)
Версия с графическим интерфейсом для MS-Windows 32 бит с поддержкой OLE
Заплатки: 1-69
Скомпилирован  mool@tororo
Огромная версия с графическим интерфейсом.  Включённые (+) и отключённые (-) особенности:
+acl                +cursorbind         +gettext/dyn        +mouse              +reltime            +timers
+arabic             +cursorshape        -hangul_input       +mouseshape         +rightleft          +title
+autocmd            +dialog_con_gui     +iconv/dyn          +multi_byte_ime/dyn +ruby/dyn           +toolbar
+balloon_eval       +diff               +insert_expand      +multi_lang         +scrollbind         +user_commands
+browse             +digraphs           +job                -mzscheme           +signs              +vertsplit
++builtin_terms     +directx            +jumplist           +netbeans_intg      +smartindent        +virtualedit
+byte_offset        -dnd                +keymap             +num64              +startuptime        +visual
[...]

PR #588, opened 2017-01-31
Fix job_start with spaces and UTF8 in path in Windows (Vim8)

slepogin:
Patch related to this issue #586 . I checked it works with spaces and UTF8 characters as well.

This is my first pull request ever please check and tell me if something wrong.
Thank you!


junegunn:
Thanks, I'll test it tonight.

junegunn:
I confirmed that the patch fixes all known issues. Thank you very much!

/cc @tjharman

PR #590, opened 2017-02-08
Introduce new command PlugOpen to browse the directory of a plugin.

alexanderjeurissen:
One of the great features of Bundler (a package manager in Ruby) is the `bundle open` command which allows me to quickly view the source of a gem.

The behaviour is quite neat after executing `bundle open` vim opens in NetRw explore mode with the gem directory is `cd`.

I wished to have similar functionality in vim where I could invoke `PlugOpen Fzf.vim` for example to open a netrw split that browses the plug directory of that plugin.

This PR does just that, it adds a new command `PlugOpen` which has autocomplete for all installed and managed Plugs, the user can supply a bang (!) to change the split behaviour from horizontal to vertical.

If you want I can add a screenrecording showing the command in action.

### DIFF
--- a/plug.vim
+++ b/plug.vim
@@ -138,6 +138,7 @@ function! s:define_commands()
   endif
   command! -nargs=* -bar -bang -complete=customlist,s:names PlugInstall call s:install(<bang>0, [<f-args>])
   command! -nargs=* -bar -bang -complete=customlist,s:names PlugUpdate  call s:update(<bang>0, [<f-args>])
+  command! -nargs=* -bar -bang -complete=customlist,s:names PlugOpen    call s:open(<bang>'',<f-args>)
   command! -nargs=0 -bar -bang PlugClean call s:clean(<bang>0)
   command! -nargs=0 -bar PlugUpgrade if s:upgrade() | execute 'source' s:esc(s:me) | endif
   command! -nargs=0 -bar PlugStatus  call s:status()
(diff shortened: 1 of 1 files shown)


junegunn:
Hi, thanks for the suggestion.

vim-plug labels itself as a "minimalist vim plugin manager" and to keep it simple and small, we encourage users to extend vim-plug using the public `g:plugs` variable instead of adding more commands and functions. Examples can be found in https://github.com/junegunn/vim-plug/wiki/extra

Likewise, the suggested feature can be implemented externally without touching the core.

```vim
function! s:complete_plug(...)
  return sort(filter(keys(g:plugs), '!stridx(v:val, a:1)'))
endfunction

command! -nargs=1 -bang -complete=customlist,s:complete_plug PlugOpen
  \ execute <bang>0 ? 'vsplit' : 'split' g:plugs[<q-args>].dir
```

And I suggest that you use the above snippet instead.

Adding a new command is not trivial. We have to add proper [test cases for it](https://github.com/junegunn/vim-plug/tree/master/test) and update documentation to explain the new feature. And usually we have to add a few configuration variables as well to meet different requirements – for example in this case, what if one wants to open it in a new tab, or in an external program like ranger, or in [a new terminal window](https://github.com/justinmk/vim-gtfo)? – which means we need even more test cases and more documentation. And still, it is not as flexible as defining a custom command using `g:plugs` as shown above.

PR #608, opened 2017-03-10
Fix crlf in clones under cygwin (#497)

christianrondeau:
This pull request is actually very similar to #497, but since it seems to have been abandoned, here you go.

I have tested under cygwin, and it works :)

In a nutshell:

1. Under Windows and Linux, there's no problem; `core.autocrlf` takes care of everything.
2. Under Cygwin, the Windows version of vim will *not* work and it will complain about `^M` everywhere.
3. I cannot configure `git` globally under cygwin, because I *want* my Windows files to be `crlf`, even though I am working in Cygwin

The solution is to have the vim plugins clone with `lf` under cygwin.

I can now SSH on my windows machine and work :)

### DIFF
--- a/plug.vim
+++ b/plug.vim
@@ -986,6 +986,10 @@ function! s:update_impl(pull, force, args) abort
   let s:clone_opt = get(g:, 'plug_shallow', 1) ?
         \ '--depth 1' . (s:git_version_requirement(1, 7, 10) ? ' --no-single-branch' : '') : ''
 
+  if has('win32unix')
+    let s:clone_opt .= ' -c core.eol=lf -c core.autocrlf=input'
+  endif
+
   " Python version requirement (>= 2.7)
   if python && !has('python3') && !ruby && !use_job && s:update.threads > 1
     redir => pyv


plug.vim
```diff
@@ -986,6 +986,10 @@ function! s:update_impl(pull, force, args) abort
   let s:clone_opt = get(g:, 'plug_shallow', 1) ?
         \ '--depth 1' . (s:git_version_requirement(1, 7, 10) ? ' --no-single-branch' : '') : ''
 
+  if(has('win32unix'))
```
junegunn: We don't use parentheses when they're not necessary. Can you remove the outer parentheses?

christianrondeau: Absolutely :)

christianrondeau:
Changed as requested and squashed!

plug.vim
```diff
@@ -986,6 +986,10 @@ function! s:update_impl(pull, force, args) abort
   let s:clone_opt = get(g:, 'plug_shallow', 1) ?
         \ '--depth 1' . (s:git_version_requirement(1, 7, 10) ? ' --no-single-branch' : '') : ''
 
+  if has('win32unix')
+    let s:clone_opt .= ' -c core.eol=lf -c core.autocrlf=input'
```
junegunn: Thanks for the update. But I'm wondering, do we need both options?

christianrondeau: Yes, I also thought that only `core.eol=lf` would suffice, but `core.autocrlf` actually overrides `core.eol`... and under Cygwin with `core.autocrlf=auto`, `crlf` is used by default. If you're curious, I can dig up the conversations on that topic I have found while doing that PR.

junegunn: I see, thanks for the clarification. I'll just go ahead and merge the patch. And yes, it would be nice if you could post a link to the conversation for future reference.

junegunn:
Merged, thanks!

christianrondeau:
https://git-scm.com/docs/gitattributes and https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration explain it:

* `Note that core.autocrlf overrides core.eol`
* `If you’re on a Linux or Mac system that uses LF line endings, then you don’t want Git to automatically convert them when you check out files; however, if a file with CRLF endings accidentally gets introduced, then you may want Git to fix it. You can tell Git to convert CRLF to LF on commit but not the other way around by setting core.autocrlf to input`
* `eol: This attribute sets a specific line-ending style to be used in the working directory. It enables end-of-line conversion without any content checks, effectively setting the text attribute. [...] Set to string value "lf: This setting forces Git to normalize line endings to LF on checkin and prevents conversion to CRLF when the file is checked out.`

But this does not make it obvious however what settings you should use on Cygwin, because it is Windows, but uses LF.

Here are a few StackOverflow references with that same conclusion:

* http://stackoverflow.com/a/9977954/154480
* http://stackoverflow.com/a/13154031/154480
* http://stackoverflow.com/a/33424884/154480

And the last but not the least... it works in my Cygwin environment :)

PR #616, opened 2017-04-11
Load plugins only once in plug#load

blueyed:
When loading 'deoplete.nvim' for the 2nd time during InsertEnter
manually, the `s:dobufread` (or `s:lod` itself) prevents it to work
properly - likely because the plugin gets resourced.

Maybe there could be a way to force this (and reload plugins always),
but by default it seems to make sense to skip already loaded plugins.

### DIFF
--- a/plug.vim
+++ b/plug.vim
@@ -447,11 +447,15 @@ function! plug#load(...)
     let s = len(unknowns) > 1 ? 's' : ''
     return s:err(printf('Unknown plugin%s: %s', s, join(unknowns, ', ')))
   end
-  for name in a:000
-    call s:lod([name], ['ftdetect', 'after/ftdetect', 'plugin', 'after/plugin'])
-  endfor
-  call s:dobufread(a:000)
-  return 1
+  let unloaded = filter(copy(a:000), '!get(s:loaded, v:val, 0)')
+  if !empty(unloaded)
+    for name in unloaded
+      call s:lod([name], ['ftdetect', 'after/ftdetect', 'plugin', 'after/plugin'])
+    endfor
+    call s:dobufread(unloaded)
+    return 1
+  end
+  return 0
 endfunction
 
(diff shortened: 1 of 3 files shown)


blueyed:
Please let me know if this makes sense in general, and I will look into the failing tests.

junegunn:
I think it makes sense and is generally favorable. There can be some users using the function to reload plugins, but since it's not clearly documented how the function behaves when the plugin is already loaded, we can change it.

Regarding the failing test cases, you may need to use `ResetPlug()` function that we inject during test for clearing `s:loaded`.

blueyed:
> return `0` when no plugin was loaded.

What do you think about returning the list of plugins that were loaded?

junegunn:
Although I like the idea, I'd like to keep the return value as boolean type as I myself already have some code depend on the return value of `plug#load()`. No need to break things when there's a compelling reason to do so.

https://github.com/junegunn/dotfiles/blob/de237c5da777e907c41f05cc55191207764b95d8/vimrc#L533-L538

plug.vim
```diff
@@ -447,11 +447,15 @@ function! plug#load(...)
...
-    call s:lod([name], ['ftdetect', 'after/ftdetect', 'plugin', 'after/plugin'])
-  endfor
-  call s:dobufread(a:000)
-  return 1
+  let unloaded = filter(copy(a:000), '!get(s:loaded, v:val, 0)')
+  if !empty(unloaded)
+    for name in unloaded
+      call s:lod([name], ['ftdetect', 'after/ftdetect', 'plugin', 'after/plugin'])
+    endfor
+    call s:dobufread(a:000)
```
junegunn: Should `a:000` be `unloaded` here?

blueyed: Yes, amended.

junegunn:
Thanks a lot!

PR #619, opened 2017-04-13
PlugStatus showed error with wildcard tag

aacn500:
:PlugStatus was reporting an invalid tag when using e.g.`Plug 'fatih/vim-go', { 'tag': '*' }`

plugstatus


junegunn:
Thanks a lot!

Issue #559, opened 2016-12-08
Install documentation for Neovim to prefer XDG_DATA_HOME?


smyrman:
I want to suggest improving the Neovim download instructions to better align with the [XDG Base directory specification](https://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html). In short, plugins, including vim-plug, are more appropriately stored as user data rather than config.

This would mean changing the dowload instructions for Neovim to:
```
curl -fLo ${XDG_DATA_HOME:-~/.local/share}/nvim/site/autoload/plug.vim --create-dirs \
    https://raw.githubusercontent.com/junegunn/vim-plug/master/plug.vim
```

In addition,
```vim
call plug#begin('~/.vim/plugged')
```
should be replaced by
```vim
call plug#begin('$XDG_DATA_HOME/nvim/plugged')
" or if XDG_DATA_HOME is not set:
call plug#begin('~/.local/shared/nvim/plugged')
```
for Neovim only.

------------------------------

<!-- Put the contents of `:version` below -->
```
VIM 0.1.7                                                                                                                                                                                                         
Build type: RelWithDebInfo                                                                                                                                                                                         
Compilation: /usr/local/Homebrew/Library/Homebrew/shims/super/clang -Wconversion -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=1 -O2 -g -DDISABLE_LOG -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=
gnu99 -Wvla -fstack-protector-strong -fdiagnostics-color=auto -DINCLUDE_GENERATED_DECLARATIONS -DHAVE_CONFIG_H -I/tmp/neovim-20161206-12093-u75jk1/neovim-0.1.7/build/config -I/tmp/neovim-20161206-12093-u75jk1/ne
ovim-0.1.7/src -I/usr/local/include -I/usr/local/include -I/usr/local/include -I/usr/local/include -I/usr/local/include -I/usr/local/opt/gettext/include -I/usr/include -I/Applications/Xcode.app/Contents/Develope
r/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.12.sdk/usr/include -I/tmp/neovim-20161206-12093-u75jk1/neovim-0.1.7/build/src/nvim/auto -I/tmp/neovim-20161206-12093-u75jk1/neovim-0.1.7/build/include         
Compiled by sindrero@sjokolade.local                                                                                                                                                                               
[...]

PR #643, opened 2017-06-14
Add .config path for installation "script"

cleac:
This is done becase latest versions of NeoVIM use this path as path where configurations
and plugins are places


junegunn:
Can you point me to the relevant documentation? The path was suggested by the maintainer of Neovim, @justinmk not too long ago. See #559

cleac:
Sorry, kinda strange thing happened and it didn't work out. I suggested the problem in that, but it seems not

Issue #197, opened 2015-03-24
On demand loading for operating system


oschrenk:
Is it possible to add syntax for plug loading based on os? For instance:

```
Plug 'glidenote/newdayone.vim', { 'os': 'darwin' } " write dayone entries
Plug 'rizzatti/dash.vim', { 'os': 'darwin' }       " search in dash
```

Or is it out of scope?

vheon:
what is wrong with:

```viml
if has('macunix')
Plug 'glidenote/newdayone.vim'
Plug 'rizzatti/dash.vim'
endif
```
??

vheon:
Similar question #137

oschrenk:
Is having one line instead of three and be visually be more pleasing a good reason?

I know of `if has('macunix')` and the  likes. Seeing #137, it seems the way for Plug is to be lean, which is not a bad way.

vheon:
> Seeing #137, it seems the way for Plug is to be lean, which is not a bad way.

Yes. I know @junegunn thinks there are too many options for the `Plug` command already.

> Is having one line instead of three and be visually be more pleasing a good reason?

Just to counterpoint this: Is having 5 `Plug`s all with `{ 'on': 'darwin' } better than close them in an `if`?? IMHO it's a lot redundant. But everyone has its taste ;)

junegunn:
@oschrenk We have concluded that the addition of such an option does not give enough benefit to justify the added complexity. You can see more discussion on the topic on #108 and #115. Thanks.

@vheon Thanks for the comment!

Issue #137, opened 2014-12-18
Plugin Sets


frace:
I do use my .vimrc on multiple machines, but often I don't need all of my listed plugins on every system. Currently I comment plugins on machines where I don't need them. But wouldn't it be nice to be able to define some kind of sets for different scenarios? I'm not quite sure how such a feature could be implemented but what came into my mind is the following idea:

```
" Choose a local name from a list of arbitrary names
" Reserved set names are: all, none
let l:vimplug_available_sets = [ 'work_laptop', 'work_workstation', 'home_laptop' ]

call plug#begin($HOME . '/.vim/plugged')

        Plug 'plugin_1', { 'set': 'work_laptop' }
        Plug 'plugin_2', { 'set': 'work_workstation' }
        Plug 'plugin_3', { 'set': 'home_laptop' }
        Plug 'plugin_5', { 'set': ['work_laptop', 'work_workstation'] }
        Plug 'plugin_6', { 'set': 'all' }
        Plug 'plugin_7', { 'set': 'none' }

call plug#end()
```

The example would allow vim-plug to install plugin_1, plugin_5 and plugin_6 on our work_laptop. Everything else would be ignored.

PR #647, opened 2017-06-27
Add operating system specific plugins and options section

herrbischoff:
Looking at #137 and #197, it could be helpful to have a more specific example in the documentation, especially one with robust OS detection.

### DIFF
--- a/README.md
+++ b/README.md
@@ -261,6 +261,22 @@ let g:fzf_install = 'yes | ./install'
 Plug 'junegunn/fzf', { 'do': g:fzf_install }
 ```
 
+### Operating system specific plugins and options
+
+While this functionality is not built into the plugin, you can easily archieve this with Vim's own commands. For example configuring the post-update hooks for `vimproc` differently on FreeBSD:
+
+```vim
+let g:os=substitute(system('uname'), '\n', '', '')
+
+if g:os ==# 'FreeBSD'
+    Plug 'Shougo/vimproc', { 'do': 'gmake' }
+else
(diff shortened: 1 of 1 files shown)


junegunn:
After those discussions, we added a section to the wiki page that deals with the issue in depth. See https://github.com/junegunn/vim-plug/wiki/faq#conditional-activation

So, instead of having another section on the README page for that, I'll try to find a nice example to add to [the current set of examples](https://github.com/junegunn/vim-plug#example).

Regarding your examples, I would advise against using `system()` function in vimrc unless it's absolutely necessary since it can increase the startup time by a few milliseconds. Instead, one should see if it's possible to do the same with `has('some_feature')` or `exists('$ENV_VARIABLE')`.

Anyway, thanks for the suggestion.

herrbischoff:
Alright, I did not see the FAQ page. Good idea with adding to the examples.

Regarding the use of `has()`, according to `:h feature-list`, Vim does not have a built-in capability to check for the OS it's running on apart from `mac`, `macunix`, `unix`, `win32`, `win32unix` and `win64`. That distinction is rather coarse for this particular use case.

junegunn:
You're right, `has()` may not suffice in those cases. In this particular case, `executable('gmake')` is a good alternative. Or I would probably just use `Plug 'Shougo/vimproc', { 'do': 'gmake || make' }`.

junegunn:
`Plug 'Shougo/vimproc', { 'do': executable('gmake') ? 'gmake' : 'make' }`.

Issue #498, opened 2016-07-01
Vim async feature for parallel update


chrisbra:
Does vim-plug use the async feature of recent Vim builds for parallel fetching of plugins?
If so, this should be documented somewhere, if not, could this be please be enabled?

```
VIM - Vi IMproved 7.4 (2013 Aug 10, compiled May 10 2016 21:39:57)
Included patches: 1-1829
Compiled by cb@256bit.org
Huge version with GTK2 GUI.  Features included (+) or not (-):
+acl             +farsi           +mouse_netterm   +tag_binary
+arabic          +file_in_path    +mouse_sgr       +tag_old_static
+autocmd         +find_in_path    -mouse_sysmouse  -tag_any_white
+balloon_eval    +float           +mouse_urxvt     -tcl
+browse          +folding         +mouse_xterm     +termguicolors
++builtin_terms  -footer          +multi_byte      +terminfo
+byte_offset     +fork()          +multi_lang      +termresponse
+channel         +gettext         -mzscheme        +textobjects
+cindent         -hangul_input    +netbeans_intg   +timers
+clientserver    +iconv           +packages        +title
+clipboard       +insert_expand   +path_extra      +toolbar
+cmdline_compl   +job             -perl            +user_commands
+cmdline_hist    +jumplist        +persistent_undo +vertsplit
+cmdline_info    +keymap          +postscript      +virtualedit
+comments        +langmap         +printer         +visual
+conceal         +libcall         +profile         +visualextra
+cryptv          +linebreak       +python/dyn      +viminfo
```

- Type:
    - [ ] Bug
    - [ ] Enhancement
    - [ ] Feature Request
    - [x] Question
- OS:
    - [x] All/Other
    - [ ] Linux
    - [ ] OS X
    - [ ] Windows
- Vim:
    - [x] Terminal Vim
    - [x] GVim
    - [ ] Neovim

junegunn:
It's not on my priority list as we already have Ruby and Python installer for Vim. Although they are not non-blocking, I don't really see the benefit of a non-blocking installer as the process usually finishes pretty quickly (`:PlugUpdate99`) and the asynchronous nature can be a problem in some contexts (see #104). Plus, it's hard to find a Vim installation without `+python` or `+ruby` these days.

[...]

PR #669, opened 2017-08-15
Check patch-8.0.87 for vim 8 job

janlazo:
Windows releases for Vim seem to always include all patches so it's safe to check only 1 patch to assume the rest are included.

Reference: https://github.com/junegunn/vim-plug/pull/666#issuecomment-322019160


junegunn:
I don't think the change is needed for non-windows environments.

janlazo:
What about patch 8.0.45 for non-windows?

junegunn:
What does patch 45 fix? I'm not sure why we should change the condition. We have confirmed that the Installer using job API works with patch 39 both on non-Windows and on Windows environment (see #534, #541). #594 not working as expected is a separate problem. Anyway, we can conditionally set up `s:vim8` depending on the environment if that's required.

```vim
let s:vim8 = exists('*job_start') && (s:is_win && has('...') || !s:is_win && has('patch-8.0.87'))
```

janlazo:
I choose those patches for #594 which seems to work on the `patch-8.0.921` release for Windows.

> What does patch 45 fix?

From the patch description in `:h patches-8`, it fixes `job_stop` for Vim 8.
Based on the [comment](https://github.com/junegunn/vim-plug/issues/498#issuecomment-254239466) for `patch-8.0.39`, I think it's worth considering.

>Anyway, we can conditionally set up s:vim8 depending on the environment if that's required. 

Simpler to require one more patch for Windows

```viml
let s:vim8 = exists('*job_start') && has('patch-8.0.39') && (!s:is_win || has('patch-8.0.87')
```

junegunn:
Thanks for clarification. I tried to separate two issues; Vim 8 installer and `s:system` using job API (#594).

1. Vim 8 installer works on Vim 8 with patch 39 or above
2. `s:system` using job API seems to require patch 45 or above

Changing the version requirement from 39 to 45 will make Vim 8 installer suddenly unavailable on a set of Vim versions (patch 39 ~ 44) even they are capable of running it. That was the issue I wanted to address. However, I guess it's not worthwhile to put more if-else statements just for that small range of versions. I'm no longer against changing the requirement.

Having said that, this should be merged with #594 (or equivalent), we don't have to merge it before it.

Issue #606, opened 2017-03-07
Avoid git submodules in Windows hanging on "Hit any key to close this window..."


dezza:
To make vim-plug completely unattended while installing plugins on Windows it should avoid the "Hit any key to close this window..." for repos with submodules by not using `!` syntax which is used in the [`s:bang(cmd, ...)`](https://github.com/junegunn/vim-plug/blob/master/plug.vim#L784-L790) function used in [`s:update_finish()`](https://github.com/junegunn/vim-plug/blob/master/plug.vim#L1080)

Instead it could use the Python [Command()](https://github.com/junegunn/vim-plug/blob/master/plug.vim#L1430) class OR
`silent !` but that is the unlikely solution - my guess is that there's a good reason it's not silent so you can verify it's not failing - so I guess the Python option is the best choice?

Please advice if this could be changed so we can avoid the `cmd.exe` window awaiting ENTER on each submodule.

I will submit a commit/pullrequest if you agree.

junegunn:
Actually `s:bang` function implements a trick for skipping "ENTER key" prompt.

https://github.com/junegunn/vim-plug/blob/master/plug.vim#L791

Seems like it's not working on Windows, right? I'd like to see a pure Vim script solution to this rather than relying on python interface which is not guaranteed to be available.

dezza:
Oh I see.

No on my official Windows Vim this still shows the prompt

```
:version
VIM - Vi IMproved 8.0 (2016 Sep 12, compiled Nov  6 2016 17:20:44)
MS-Windows 32-bit GUI version with OLE support
Included patches: 1-69
Compiled by mool@tororo        
```

I will be willing to experiment with fixing this later and I agree I also do as much as possible in VimScript 👍

nzbart:
@dezza did you make much progress on this? It would make unattended installations a lot smoother...

dezza:
@nzbart I opted for another solution (https://github.com/airblade/voom)

I guess it's pretty easy once you find the offending line.

nzbart:
@dezza thanks; something simple like voom sounds really good... unfortunately, it's not cross-platform, which I need it to be.

dezza:
It is for me since I utilize WSL for Windows (shell) and mac already has shell.

janlazo:
Is this for GVim only?

janlazo:
@junegunn I can fix this but I need `fzf#shellescape` to replace `s:shellesc` for `s:with_cd`.

[...]

Issue #668, opened 2017-08-15
shellescape comment in s:bang


janlazo:
Does porting the `fzf#shellescape` fix the escaping issue here for Windows?

```viml
function! s:bang(cmd, ...)
  try
    let [sh, shrd] = s:chsh(a:0)
    " FIXME: Escaping is incomplete. We could use shellescape with eval,
    "        but it won't work on Windows.
    let cmd = a:0 ? s:with_cd(a:cmd, a:1) : a:cmd
    let g:_plug_bang = '!'.escape(cmd, '#!%')
    execute "normal! :execute g:_plug_bang\<cr>\<cr>"
  finally
    unlet g:_plug_bang
    let [&shell, &shellredir] = [sh, shrd]
  endtry
  return v:shell_error ? 'Exit status: ' . v:shell_error : ''
endfunction
```

- Type:
    - [ ] Bug
    - [X] Enhancement
    - [ ] Feature Request
    - [X] Question
- OS:
    - [ ] All/Other
    - [ ] Linux
    - [ ] OS X
    - [X] Windows
- Vim:
    - [X] Terminal Vim
    - [X] GVim
    - [X] Neovim

junegunn:
Possibly. However, no one has ever reported an issue regarding the code. I think it's because no one  really uses complex `do` option value that can cause escaping issues in reality. So I haven't been actively trying to "fix" the code.

janlazo:
It can't handle filepaths with spaces in Windows.

janlazo:
Is the `FIXME` comment still valid?

Issue #635, opened 2017-05-22
Shellxquote settings breaking things cryptically


Taverius:
I had an old `shellxquote=""` from way back when it was required to have such odd things for anything at all to work on windows, when dinosaurs roamed the earth.

It never caused issues and fixed some, so in my .vimrc it stayed, until today.

Today I decided to drop NeoBundle for Plug, and I was getting cryptic error messages when I attempted to run `PlugUpdate` to see if everything was working:

![plug_breaks](https://cloud.githubusercontent.com/assets/354219/26308011/a8961226-3ef7-11e7-84f4-8333b91d993c.png)

Turns out my `shellxquote` setting was causing it.

A quick look says that Windows defaults are sane now, so that's good at least.

Can you add a warning, or a faq item, or a workaround for that?

------------------------------

```
VIM - Vi IMproved 8.0 (2016 Sep 12, compiled May 17 2017 21:37:07)
MS-Windows 64-bit console version
Included patches: 1-599
Compiled by hp@HP-PC
Huge version without GUI.  Features included (+) or not (-):
+acl             +ex_extra        -mouseshape      +tag_old_static
+arabic          +extra_search    +multi_byte      -tag_any_white
+autocmd         +farsi           +multi_lang      +tcl/dyn
-balloon_eval    +file_in_path    +mzscheme/dyn    +termguicolors
-browse          +find_in_path    -netbeans_intg   -tgetent
++builtin_terms  +float           +num64           -termresponse
+byte_offset     +folding         +packages        +textobjects
+channel         -footer          +path_extra      +timers
+cindent         +gettext/dyn     +perl/dyn        +title
+clientserver    -hangul_input    +persistent_undo -toolbar
+clipboard       +iconv/dyn       -postscript      +user_commands
+cmdline_compl   +insert_expand   +printer         +vertsplit
+cmdline_hist    +job             +profile         +virtualedit
+cmdline_info    +jumplist        +python/dyn      +visual
+comments        +keymap          +python3/dyn     +visualextra
+conceal         +lambda          +quickfix        +viminfo
+cryptv          +langmap         +reltime         +vreplace
+cscope          +libcall         +rightleft       +wildignore
+cursorbind      +linebreak       +ruby/dyn        +wildmenu
+cursorshape     +lispindent      +scrollbind      +windows
+dialog_con      +listcmds        +signs           +writebackup
+diff            +localmap        +smartindent     -xfontset
+digraphs        +lua/dyn         +startuptime     -xim
[...]

Issue #539, opened 2016-10-17
PlugInstall breaks if Vim started from Git-for-Windows (MSYS2)


go2null:
`PlugInstall` failure when `vim` is started from `MSYS2`

```
x vim-ruby:
    /usr/bin/bash: line 0: cd: /d: No such file or directory
    PlugClean required.
```

---

```
VIM - Vi IMproved 7.4 (2013 Aug 10, compiled Aug 10 2013 14:38:33)
MS-Windows 32-bit GUI version with OLE support
Compiled by mool@tororo
Big version with GUI.  Features included (+) or not (-):
+arabic             +ex_extra           +mouseshape         +syntax
+autocmd            +extra_search       +multi_byte_ime/dyn +tag_binary
+balloon_eval       +farsi              +multi_lang         +tag_old_static
+browse             +file_in_path       -mzscheme           -tag_any_white
++builtin_terms     +find_in_path       +netbeans_intg      +tcl/dyn
+byte_offset        +float              +ole                -tgetent
+cindent            +folding            +path_extra         -termresponse
+clientserver       -footer             +perl/dyn           +textobjects
+clipboard          +gettext/dyn        +persistent_undo    +title
+cmdline_compl      -hangul_input       -postscript         +toolbar
+cmdline_hist       +iconv/dyn          +printer            +user_commands
+cmdline_info       +insert_expand      -profile            +vertsplit
+comments           +jumplist           +python/dyn         +virtualedit
+conceal            +keymap             +python3/dyn        +visual
+cryptv             +langmap            +quickfix           +visualextra
+cscope             +libcall            +reltime            +viminfo
+cursorbind         +linebreak          +rightleft          +vreplace
+cursorshape        +lispindent         +ruby/dyn           +wildignore
+dialog_con_gui     +listcmds           +scrollbind         +wildmenu
+diff               +localmap           +signs              +windows
+digraphs           -lua                +smartindent        +writebackup
-dnd                +menu               -sniff              -xfontset
-ebcdic             +mksession          +startuptime        -xim
+emacs_tags         +modify_fname       +statusline         -xterm_save
+eval               +mouse              -sun_workshop       +xpm_w32
   system vimrc file: "$VIM\vimrc"
     user vimrc file: "$HOME\_vimrc"
 2nd user vimrc file: "$HOME\vimfiles\vimrc"
 3rd user vimrc file: "$VIM\_vimrc"
      user exrc file: "$HOME\_exrc"
[...]

PR #676, opened 2017-09-14
Port fzf#shellescape to fix escaping issues in Windows

janlazo:
Initial work to fix all escaping issues for Vim and Neovim in Windows. `tempname` is necessary to bypass Vim's regular shellescaping in both `s:system` and `s:bang` like in Vim plugin of fzf because regular Vim seems to assume that the shell options are unchanged in Windows even if all but `shell` are unset. It's possible to not require batchfiles but it means double-escaping (ie. `fzf#shellescape(fzf#shellescape(cmd))` ) for Vim and single-escaping for Neovim with the same values for shell options.

### DIFF
--- a/plug.vim
+++ b/plug.vim
@@ -782,8 +782,10 @@ function! s:assign_name()
 endfunction
 
 function! s:chsh(swap)
-  let prev = [&shell, &shellredir]
-  if !s:is_win && a:swap
+  let prev = [&shell, &shellcmdflag, &shellredir]
+  if s:is_win
+    set shell=cmd.exe shellcmdflag=/c shellredir=>%s\ 2>&1
+  elseif a:swap
     set shell=sh shellredir=>%s\ 2>&1
   endif
   return prev
@@ -791,15 +793,23 @@ endfunction
 
 function! s:bang(cmd, ...)
   try
-    let [sh, shrd] = s:chsh(a:0)
+    let [sh, shellcmdflag, shrd] = s:chsh(a:0)
     " FIXME: Escaping is incomplete. We could use shellescape with eval,
     "        but it won't work on Windows.
     let cmd = a:0 ? s:with_cd(a:cmd, a:1) : a:cmd
+    if s:is_win
+      let batchfile = tempname().'.bat'
+      call writefile(['@echo off', cmd], batchfile)
+      let cmd = batchfile
+    endif
     let g:_plug_bang = '!'.escape(cmd, '#!%')
     execute "normal! :execute g:_plug_bang\<cr>\<cr>"
   finally
     unlet g:_plug_bang
-    let [&shell, &shellredir] = [sh, shrd]
+    let [&shell, &shellcmdflag, &shellredir] = [sh, shellcmdflag, shrd]
+    if s:is_win
+      call delete(batchfile)
+    endif
   endtry
   return v:shell_error ? 'Exit status: ' . v:shell_error : ''
 endfunction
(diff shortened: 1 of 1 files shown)


plug.vim
```diff
@@ -2316,10 +2345,19 @@ function! s:preview_commit()
...
-    let [sh, shrd] = s:chsh(1)
-    execute 'silent %!cd' s:shellesc(g:plugs[name].dir) '&& git show --no-color --pretty=medium' sha
+    let [sh, shellcmdflag, shrd] = s:chsh(1)
+    let cmd = 'cd '.s:shellesc(g:plugs[name].dir).' && git show --no-color --pretty=medium '.sha
+    if s:is_win
+      let batchfile = tempname().'.bat'
+      call writefile(['@echo off', cmd], batchfile)
+      let cmd = batchfile
+    endif
+    execute 'silent %!' cmd
```
janlazo: This looks like `s:bang` now but `%#!` are not escaped.

janlazo:
`s:spawn` needs `tempname` for the jobs for the same reasons on `system` and `:!`, especially in regular Vim 8.

janlazo:
@junegunn Do I have to update the python and ruby jobs?

junegunn:
> Do I have to update the python and ruby jobs?

No, we don't want to spend more of our time on those legacy installers.

So this fixes #606, right? It's not clear to me though how using a batchfile instead of a command string suppresses "Enter key" prompt. Can you explain?

janlazo:
This targets #635, #668, #539 ('shell' points to `$SHELL`), Case 1 in https://github.com/junegunn/vim-plug/wiki/faq#windows-system-error-e484.

For 606, appending `silent` to the bang command in `s:bang` (see https://github.com/junegunn/vim-plug/issues/606#issuecomment-326469010) is enough. I don't know if this is required for MacVim as well.

junegunn:
> This PR targets #635, #668, #539 

Thanks for the clarification. I completely lost track of those issues, largely because I'm not a Windows user.

I just reviewed the code, although it's unfortunate that we have to introduce extra code branches here and there for Windows, the change looks good to me. Is there anything else do you want to address?

janlazo:
>  Can you update the commit message with `Close #xxx` lines so that those issues are automatically closed when this is merged?

I'll rebase when I think it's ready.

> Is there anything else you want to address?

I prefer vim-plug to use `cmd.exe` only in Windows so the side effects of setting `SHELL` before Vim runs should be reverted for `system` to work properly.

janlazo:
@junegunn do you want the fix for #606 in here or in a separate PR?

janlazo:
@junegunn Unlike fzf's Vim plugin, `shell` is not set at the start so `s:shellesc` may not work as expected. To guarantee that it will work in the batchfile, I think it's better to make it non-configurable.

```viml
if has('win32')
  function s:shellesc(arg)
    return s:shellesc_cmd(a:arg)
  endfunction
else
  function s:shellesc(arg)
    return shellescape(a:arg)
  endfunction
endif
```

junegunn:
> do you want the fix for #606 in here or in a separate PR?

As long as it's on a separate commit, I'm fine with either :) By the way, we are going to apply `silent` only on GVim, right? As you did in https://github.com/junegunn/vim-plug/issues/606#issuecomment-326469010. On terminal Vim, it messes up the screen and requires additional `:redraw!`.

> do you want to keep the old s:shellesc as fallback instead of shellescape?

No, you properly handled escaping issues on Windows, so I guess that crude solution is no longer needed.

junegunn:
Reviewed the code and it looks good to me. Thanks, again. I see that you rebased your commits, ready for merge?

janlazo:
> As long as it's on a separate commit, I'm fine with either :) By the way, we are going to apply `silent` only on GVim, right?

Yes. I'll open a separate PR for #606 after the merge.

> ready for merge?

Yes for Windows unless you want to handle msysgit or cygwin as well.

junegunn:
Thanks, I really appreciate your contribution.

janlazo:
@junegunn Do you know why some `s:system` calls use `s:esc` for tags, branches instead of `s:shellesc`?

I got https://github.com/junegunn/vim-plug/commit/eb47183af2374f278b776f369c8cecb3c9441f49 from the blame. I don't use tags or branches so I didn't notice while working on this PR.

PR #678, opened 2017-09-19
Prepend 'silent' to ':!' to avoid hit-enter prompt in GVim on Windows

### LINKED ISSUE #606: Avoid git submodules in Windows hanging on "Hit any key to close this window..."

janlazo:
Close #606

### DIFF
--- a/plug.vim
+++ b/plug.vim
@@ -802,7 +802,7 @@ function! s:bang(cmd, ...)
       call writefile(['@echo off', cmd], batchfile)
       let cmd = batchfile
     endif
-    let g:_plug_bang = '!'.escape(cmd, '#!%')
+    let g:_plug_bang = (s:is_win && has('gui_running') ? 'silent ' : '').'!'.escape(cmd, '#!%')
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks!

PR #691, opened 2017-10-11
Do not trigger filetypeindent/filetypeplugin autocmds by default

blueyed:
This is not necessary if `filetype plugin indent on` was not used before
`plug#end()`, since then the `FileType` autocmds from there will come
after vim-plug's.

This will issue a warning, and makes handling of this conditional.

This could use `filetype plugin/indent off` to work around this (similar
to the `filetype off` being used), but `runtime/indoff.vim` and
`runtime/ftplugof.vim` will only empty the augroups, and not remove
them.

Fixing the user's config is the best solution anyway, so I think a
warning is good.


blueyed:
Rebased and added queuing of autocommands before VimEnter in general (https://github.com/blueyed/vim-plug/commit/00d3f0ada65c0ce84263f9beb5d8daae3989aa7d, https://github.com/vim/vim/issues/2810).

janlazo:
@blueyed Can you rebase again if it's useful to have?

PR #694, opened 2017-10-14
Use s:path with s:rtp always

blueyed:
Having a trailing slash with &rtp entries is problematic when removing
them later: if loading on demand is used, s:reorg_rtp might fail to
remove the previous runtime paths.

Test case has been using maktaba (which triggers unsetting s:middle in
s:reorg_rtp), but because of the trailing slashes being used, the
previous rtp entries were not removed.

### DIFF
--- a/plug.vim
+++ b/plug.vim
@@ -334,11 +334,11 @@ function! s:progress_opt(base)
         \ s:git_version_requirement(1, 7, 1) ? '--progress' : ''
 endfunction
 
-if s:is_win
-  function! s:rtp(spec)
-    return s:path(a:spec.dir . get(a:spec, 'rtp', ''))
-  endfunction
+function! s:rtp(spec)
+  return s:path(a:spec.dir . get(a:spec, 'rtp', ''))
+endfunction
 
+if s:is_win
   function! s:path(path)
     return s:trim(substitute(a:path, '/', '\', 'g'))
   endfunction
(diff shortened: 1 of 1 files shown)


junegunn:
Can you provide the steps to reproduce the case?

blueyed:
I've noticed it when using maktaba, which triggers changing the &rtp:
```vim
    Plug 'google/vim-maktaba'
    Plug 'google/vim-coverage'

    Plug 'Vimjas/vim-python-pep8-indent', {'for': ['python']}
```

When opening some Python file the &rtp should get doubled, having entries with and without a trailing slash.

It looked a bit like at some point Vim/Neovim would normalize the trailing slashes away.

blueyed:
Rebased.
Just noticed this again.
Without this patch/PR I get 188 entries in my &rtp, using this fix it's only 76.  A set of entries are contained therein 3 times (first with trailing slash, then without, and then with a trailing slash again).

Both start with `/home/user/.config/nvim`, and then the version without this patch has entries ending with a slash up to `/home/user/.dotfiles/vim/plugged/vim-coverage`, and then again a list without slashes after `/home/user/.dotfiles/vim/plugged/vim-maktaba` (making those appear 3 times).

janlazo:
@blueyed Can you verify if the problem exists? I'm interested in a smaller `&rtp`. Can you add a test as well?

blueyed:
@janlazo 
Have you checked your rtp with and without this patch?
(rebased for now)

janlazo:
Works for me.

PR #705, opened 2017-11-15
Vim plugin with documentation

HugoNikanor:
I felt a lack of help inside vim. So I put most of the vim-plug README into a vim help page.
That help page is here included in a plugin, for easy installation.

### DIFF
--- a/vim-plug-doc/README.md
+++ b/vim-plug-doc/README.md
@@ -1,0 +1,12 @@
+vim-plug help pages!
+===================
+Small vim plugin which gives access to vim-plug usage from
+within vim.
+
+Installation
+------------
+Add the following to your list of pluggins:
+
+```vim
+:Plug 'junegunn/vim-plug', {'rtp': 'vim-plug-doc'}
+```
--- a/vim-plug-doc/doc/vim-plug.txt
+++ b/vim-plug-doc/doc/vim-plug.txt
@@ -1,3 +1,155 @@
 *vim-plug*
+       _                       _             ~
+__   _(_)_ __ ___        _ __ | |_   _  __ _ ~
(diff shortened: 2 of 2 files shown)


partounian:
Thanks so much! I always have had trouble finding the 

> `Plug` options                                         *vim-plug-plug-options* 

Also just found that typo haha

junegunn:
Thanks for your work. Actually, I wrote a Markdown-to-Vimdoc converter, so I'll just add the file myself.

I've been reluctant to add Vim help file as I didn't want to maintain two copies of the document (you can find a couple of old discussion on this) but the feature set of vim-plug is now virtually frozen, I think I can handle that now.

Will simply add it to `doc` directory, as it should feel a bit strange to the users who are not familiar with the concept of `&runtimepath` that they have to specify `rtp` option. So it becomes:

```vim
" (Optional) If you want to register Vim help file for vim-plug 
Plug 'junegunn/vim-plug'
```

HugoNikanor:
Thank you. Having the plugin directly at the root might be simpler.

Is the markdown converter set to auto-run when the README changes? or do
you still have to do that manually?
On Sun, Nov 26, 2017 at 04:00:13PM +0000, Junegunn Choi wrote:
> Closed #705 via 96375cb7d30b7caf2a9417799ff509b59b43a627.
> 
> -- 
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub:
> https://github.com/junegunn/vim-plug/pull/705#event-1358212778

-- 
Hugo Hörnquist

junegunn:
Manually. I can set up a pre-commit git hook to verify that the file is updated accordingly, but it's still a manual process.

HugoNikanor:
At least the hook is there. I was just worried that the vim-help would
get horribly outdated.
-- 
Hugo Hörnquist

PR #707, opened 2017-11-21
Open plugin's readme file

Freedzone:
No dependencies required. 

Tested on all plugins in my .vimrc and different lines with ' to exclude errors.
Plug 'kien/ctrlp.vim'
Plug 'scrooloose/nerdtree'
Plug 'scrooloose/syntastic'
Plug 'majutsushi/tagbar'
Plug 'davidhalter/jedi-vim'
....
Plug 'romainl/Apprentice'
Plug 'junegunn/seoul256.vim'

### DIFF
--- a/plug.vim
+++ b/plug.vim
@@ -146,6 +146,7 @@ function! s:define_commands()
   command! -nargs=0 -bar PlugStatus  call s:status()
   command! -nargs=0 -bar PlugDiff    call s:diff()
   command! -nargs=? -bar -bang -complete=file PlugSnapshot call s:snapshot(<bang>0, <f-args>)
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks for your interest in the project, but I'm currently not interested in expanding the feature set of vim-plug. vim-plug exposes its core data structure via `g:plugs` so one can easily write external scripts using it as suggested in https://github.com/junegunn/vim-plug/wiki/extra. Those features can reside in one's configuration file, or in an external git repository as a bundle.

I'll see if I can learn something from your version of PlugHelp and update the example on the wiki page. Closing, thanks.

PR #711, opened 2017-11-25
Add another install way

wfxr:
Add another install way which vim will download and install the `vim-plug` automatically. Nothing will happen if it was already installed.
The benefit of this approach is that we don't need to run the install command manually. When restoring configurations on a new machine, only thing to do is copy the `vimrc` file or make a symbol link. :smiley:

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,9 +31,20 @@ and put it in the "autoload" directory.
 
 ###### Unix
 
+- Terminal way:
+
 ```sh
 curl -fLo ~/.vim/autoload/plug.vim --create-dirs \
     https://raw.githubusercontent.com/junegunn/vim-plug/master/plug.vim
+```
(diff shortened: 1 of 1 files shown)


junegunn:
Hi, the method is currently featured on the wiki page: https://github.com/junegunn/vim-plug/wiki/faq

I certainly think it's nice to make one's Vim configuration file self-contained, but I would like to present this tip to the users as an "idea" to enhance the basic installation method than something they can blindly copy & paste the code without thinking, as we don't want to repeat it three more times for Neovim and Windows.

I think the wiki page should be split into two pages, faq and tips, and I'll add the link to tips page on the README page so that it's more accessible.

wfxr:
:smile:  Agree with your choice.

PR #714, opened 2017-12-15
Corrects neovim installation directory

ecbrodie:
This is to match the more correct location for the `autoload` directory, so that it matches the [instructions on the Wiki](https://travis-ci.org/nulogy/Gorgon/builds).

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ $uri = 'https://raw.githubusercontent.com/junegunn/vim-plug/master/plug.vim'
 ###### Unix
 
 ```sh
-curl -fLo ~/.local/share/nvim/site/autoload/plug.vim --create-dirs \
+curl -fLo ~/.vim/autoload/plug.vim --create-dirs \
     https://raw.githubusercontent.com/junegunn/vim-plug/master/plug.vim
 ```
 


ecbrodie:
Note that the Travis builds are currently failing right now, but I highly believe that it is an unrelated problem to my PR, as I am only changing the README.

junegunn:
I think you're mistaken, the path was suggested by the maintainer of Neovim. See the discussion in #559.

ecbrodie:
Okay @junegunn , that's fair. But then, if the README indeed has the correct information, should the Wiki page then match that of the README?

I don't mind making that change, in that case.

And I don't mind being corrected, thank you for the updated information :)

junegunn:
The information is correct for regular Vim. And we don't want to repeat ourselves here and there for every alternative implementation of Vim. When one chooses to use an alternative rather than the base implementation that the majority of the users use, one should be aware of its differences.

junegunn:
By the way, whice wiki page are you referring to? The link you posted above is pointing to a travi-ci page. I'll update the page if it contains outdated information.

ecbrodie:
@junegunn My bad on the incorrect link! Friday night, less caffeine in the system...inexcusable!

I have corrected my original PR description. Pasting it here too: https://github.com/junegunn/vim-plug/wiki/tips. Check out the *Automatic Installation* section.

Issue #719, opened 2018-01-05
Wildcard tag doesn't play well with `:PlugUpdate`


autozimu:
(I was under the impression that this plugin supports wildcard tag releases properly, so correct me if I am wrong.)

`:PlugUpdate` seems won't fetch latest tags when the plugin has property of wildcard tag.

With this minimal vimrc
```
call plug#begin('~/.vim/plugged')
Plug 'autozimu/plug-wildcard', {'tag': 'release-*'}
call plug#end()
```

1. Run `:PlugInstall`. This will install the plugin properly.
1. Make a newer release in the repo, say `release-0.2`.
1. Run `:PlugUpdate`.

At this point, the plugin is not updated, while it is expected to update to `release-0.2`.

(This affects https://github.com/autozimu/LanguageClient-neovim/issues/239)

------------------------------

Neovim version
```
NVIM v0.2.2
Build type: Release
LuaJIT 2.0.5
Compilation: /usr/bin/cc -Wconversion -DNVIM_MSGPACK_HAS_FLOAT32 -DNVIM_UNIBI_HAS_VAR_FROM -O2 -DNDEBUG -DMIN_LOG_LEVEL=3 -Wall -Wextra -pedantic -Wno-unused-parameter -Wstrict-prototypes -std=gnu99 -Wvla -fstack-protector --param ssp-buffer-size=4 -Wno-array-bounds -DINCLUDE_GENERATED_DECLARATIONS -D_GNU_SOURCE -I/opt/neovim/build/config -I/opt/neovim/src -I/opt/neovim/.deps/usr/include -I/opt/neovim/.deps/usr/include -I/opt/neovim/.deps/usr/include -I/opt/neovim/.deps/usr/include -I/opt/neovim/.deps/usr/include -I/opt/neovim/.deps/usr/include -I/nix/store/s96g9b8rl9ab13v1by7n661h0hzpbgsw-glibc-2.25-49-dev/include -I/opt/neovim/build/src/nvim/auto -I/opt/neovim/build/include

Features: +acl +iconv +jemalloc +tui
See ":help feature-compile"

   system vimrc file: "$VIM/sysinit.vim"
  fall-back for $VIM: "/usr/local/share/nvim"

Run :checkhealth for more info
```

- Type:
    - [x] Bug
    - [ ] Enhancement
    - [ ] Feature Request
    - [ ] Question
- OS:
    - [ ] All/Other
    - [x] Linux
    - [ ] OS X
    - [ ] Windows
[...]

PR #720, opened 2018-01-06
Fetch all tags if there is `tag` spec.

autozimu:
Ref #719.


autozimu:
Need more time to look into the tests.

blueyed:
Test failures might be unrelated: https://github.com/junegunn/vim-plug/pull/724.

But likely could use a test itself.

janlazo:
Build failed in a rebased branch.
https://travis-ci.org/junegunn/vim-plug/builds/604736561

matthewarmand:
I'll preface this by saying this is a bit of an odd case, and if `vim-plug` doesn't want to support this case, that's understandable to me (but it would be good to get a statement from maintainers one way or the other).

Black (a python code formatter) does a somewhat interesting thing with tags where, in addition to having version tags, there's a `stable` tag which gets moved around to whatever the latest release is. Their editor installation docs instruct to use this with `vim-plug` (yes, I know, they say `branch` and should probably say `tag`, but that's the second issue here). By default, `git fetch --tags` and `git pull --tags` won't clobber existing tags if they've changed, but in this case that's exactly what needs to happen. (Related upstream issue: https://github.com/psf/black/issues/2503) This can be fixed by using the `force` flag. So, for example `git fetch/pull -q --tags --force`.

So, if when fetching the tags (like in this PR) `--force` was also used, it would satisfy projects which have a `stable` or `latest` tag that moves around with new releases. Then Black's docs could be changed to something like `Plug 'psf/black, { 'tag': 'stable' }` and  `vim-plug` would work for getting the latest stable version of the plugin.

Is this something maintainers would be ok with? I mention here because tag-fetching seems mostly done already here (failing tests notwithstanding)

junegunn:
@matthewarmand 

> there's a stable tag which gets moved around

That's a strange practice. Not sure if we should enforce `--force` for an exceptional case.

But since they also publish version tags, you can do something like this:

```vim
" *.* to match version numbers
Plug 'psf/black', { 'tag': '*.*' }
```

```
Updated. Elapsed time: 2.342714 sec.
[=]

- Finishing ... Done!
- Latest tag for *.* -> 22.3.0 (black)
- black: Resolving deltas: 100% (5459/5459), done.
```

matthewarmand:
Thanks @junegunn, I agree that's a much better solution. I'll open a PR soon over there suggesting that change in their docs (I had another addition I wanted to make anyway).

PR #723, opened 2018-01-14
Travis: rename vim72 to vim74

blueyed:


stappersg:
In https://travis-ci.org/junegunn/vim-plug/jobs/463001454 at line 452 is
`Setting up vim-nox (2:7.4.052-1ubuntu3.1) ...`  ( the **7.4** )

So it makes sense to do the `s/vim72/vim74/`

junegunn:
Thanks.

It's unfortunate that we don't have an easy way to test vim-plug against Vim 7.2 as Travis CI no longer seems to support Lucid environment.

blueyed:
https://github.com/tweekmonster/vim-testbed/ could be used to test against different Vims in a Docker image, but is not trivial to setup.

PR #731, opened 2018-01-27
Add a note to the installation section about following the usage section

axelson:
I spent an inordinate amount of time trying to figure out why after installation I could not simply follow a simple installation instruction like

```
    Add Plug 'iCyMind/NeoSolarized' to your init.vim or .vimrc file.
    Run :PlugInstall after resourcing/relaunching.
```

And then I saw the "Usage" section. But initially I skipped over that since I was done installing vim-plug. (also I think "Configuration" would make more sense than "Usage").

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,8 @@ A minimalist Vim plugin manager.
 [Download plug.vim](https://raw.githubusercontent.com/junegunn/vim-plug/master/plug.vim)
 and put it in the "autoload" directory.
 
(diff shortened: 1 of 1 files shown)


blueyed:
Maybe move `### Getting Help` down, and rename `Usage` to `Usage / Configuration`?

See also #737.

janlazo:
README, https://github.com/junegunn/vim-plug/blob/master/doc/plug.txt, and the wiki should be sufficent. Read the documentation next time.

PR #754, opened 2018-05-05
Fix Windows Batch file format

watagashi:
`writefile()` always output LF without CR each lines.
But batch file on Windows needs CR and LF, at end of lines.

And if the path of home directory contains non-ASCII characters like Japanese username (e.g. `C:\Users\太郎`), batch file without CR can't be executed correctly.


watagashi:
I'm sorry. There was some mistakes in this PR.

I will resubmit.

PR #755, opened 2018-05-05
Fix Windows batch file format

watagashi:
`writefile()` always output LF without CR each lines.
But batch file on Windows needs CR and LF, at end of lines.

And if the path of home directory contains non-ASCII characters like Japanese username (e.g. `C:\Users\太郎`), batch file without CR can't be executed correctly.

### DIFF
--- a/plug.vim
+++ b/plug.vim
@@ -799,7 +799,7 @@ function! s:bang(cmd, ...)
     let cmd = a:0 ? s:with_cd(a:cmd, a:1) : a:cmd
     if s:is_win
       let batchfile = tempname().'.bat'
-      call writefile(['@echo off', cmd], batchfile)
+      call writefile(["@echo off\r", cmd . "\r"], batchfile)
       let cmd = batchfile
     endif
     let g:_plug_bang = (s:is_win && has('gui_running') ? 'silent ' : '').'!'.escape(cmd, '#!%')
@@ -1196,7 +1196,7 @@ function! s:spawn(name, cmd, opts)
   let s:jobs[a:name] = job
   let cmd = has_key(a:opts, 'dir') ? s:with_cd(a:cmd, a:opts.dir) : a:cmd
   if !empty(job.batchfile)
-    call writefile(['@echo off', cmd], job.batchfile)
+    call writefile(["@echo off\r", cmd . "\r"], job.batchfile)
     let cmd = job.batchfile
   endif
   let argv = add(s:is_win ? ['cmd', '/c'] : ['sh', '-c'], cmd)
@@ -2023,7 +2023,7 @@ function! s:system(cmd, ...)
     let cmd = a:0 > 0 ? s:with_cd(a:cmd, a:1) : a:cmd
     if s:is_win
       let batchfile = tempname().'.bat'
-      call writefile(['@echo off', cmd], batchfile)
+      call writefile(["@echo off\r", cmd . "\r"], batchfile)
       let cmd = batchfile
     endif
     return system(s:is_win ? '('.cmd.')' : cmd)
@@ -2357,7 +2357,7 @@ function! s:preview_commit()
     let cmd = 'cd '.s:shellesc(g:plugs[name].dir).' && git show --no-color --pretty=medium '.sha
     if s:is_win
       let batchfile = tempname().'.bat'
-      call writefile(['@echo off', cmd], batchfile)
+      call writefile(["@echo off\r", cmd . "\r"], batchfile)
       let cmd = batchfile
     endif
     execute 'silent %!' cmd


janlazo:
Is this for `chcp 932` or `set encoding=cp932`?

plug.vim
```diff
@@ -799,7 +799,7 @@ function! s:bang(cmd, ...)
     let cmd = a:0 ? s:with_cd(a:cmd, a:1) : a:cmd
     if s:is_win
       let batchfile = tempname().'.bat'
-      call writefile(['@echo off', cmd], batchfile)
+      call writefile(["@echo off\r", cmd . "\r"], batchfile)
```
janlazo: Restrict to Windows only with `(has('win32') ? "\r" : '')`.

watagashi:
@janlazo Thanks for comments.

> Is this for `chcp 932` or `set encoding=cp932?`

No. Regardless code page or encoding, the end of the line of the batch file should be CRLF.

> If the path for the batchfile is escaped with `fzf#shellescape`, does it work without `\r`?

Does it mean like this patch?
https://gist.github.com/watagashi/0c24611807c71104ab2c01e72fbc93a3

This patch is not work.

plug.vim
watagashi: `has('win32')` is unnecessary because `if s:is_win` means it.

janlazo:
It works with LF on mine and I'm using 8.0.1790. Which vim release and Windows version do you use?

plug.vim
janlazo: You're right. I forgot about my changes.

watagashi:
@janlazo Windows command prompt on Japanese environment is not used Unicode, it is used CP932 (Shift_JIS).

The following reproduction procedure is not not required vim-plugged.

path with multibyte characters

1.  Open command prompt (`cmd.exe`)

    ```cmd
    mkdir c:\temp
    md c:\temp\太郎\vimfiles\plugged
    cd /d c:\temp\太郎\vimfiles\plugged
    git clone https://github.com/junegunn/vim-emoji.git
    cd /d c:\temp\太郎
    vim -u NORC crlf.bat
    ```

2.  Create batch file.
    (It's simulated `s:system()`)

    ```cmd
    @echo off
    cd /d ^"C:\temp\太郎\vimfiles\plugged\vim-emoji\\^" && git rev-parse --abbrev-ref HEAD 2>&1 && git config -f .git/config remote.origin.url
    ```

3.  Write files with `++ff=unix` and `++ff=dos`

    ```vim
    :w ++ff=dos " simulate writefile() with `\r`
    :w ++ff=unix lf.bat " simulate writefile() without `\r`
    ```

4.  Run batch files via Vim

    ```vim
    :echomsg system('c:\temp\太郎\crlf.bat')
    " master^@https://github.com/junegunn/vim-emoji.git^@
    :echomsg system('c:\temp\太郎\lf.bat')
    " The system cannot find the path specified.^@
    ```

    Error occurred.

path without multibyte characters

1.  Rename `C:\temp\太郎` to `C:\temp\taro`

    ```cmd
    cd C:\temp
    move C:\temp\太郎 C:\temp\taro
    cd C:\temp\taro
    del *.bat
    vim -u NORC crlf.bat
    ```

2.  create batch file.
    (It's simulated `s:system()`)

    ```cmd
    @echo off
    cd /d ^"C:\temp\taro\vimfiles\plugged\vim-emoji\\^" && git rev-parse --abbrev-ref HEAD 2>&1 && git config -f .git/config remote.origin.url
    ```

3.  Write files with `++ff=unix` and `++ff=dos`

    ```vim
    :w ++ff=dos " simulate writefile() with `\r`
    :w ++ff=unix lf.bat " simulate writefile() without `\r`
    ```

4.  Run batch files via Vim

    ```vim
    :echomsg system('c:\temp\taro\crlf.bat')
    " master^@https://github.com/junegunn/vim-emoji.git^@
    :echomsg system('c:\temp\taro\lf.bat')
    " master^@https://github.com/junegunn/vim-emoji.git^@
    ```

    No error.

***

My enviroments on Windows are

-   Windows 7 (Version 6.1.7601)
    -   Vim 8.0.1813
    -   Vim 8.0.596
-   Windows 10 (Version 10.0.17134.48)
    -   Vim 8.0.596

janlazo:
@watagashi Thanks for the explanation. I see the problem now but I have trouble reproducing the solution.

I created a directory (ie. `%USERPROFILE%\tmp\太郎`) with the multibyte characters you used and created a batchfile containing `echo a`. I can't execute the batchfile using the absolute filepath so I have to `cd` into the directory first and then exec the batchfile.

I then tried to write a batchfile via vim with the following script on GVim for `set enc=utf-8`. I couldn't run the same script in terminal Vim because of `fileencoding` (?). All three test cases pass on GVim.

```vim
" Keep the prefixed space because Vim hijacked it for !start
let cmd = ' start /wait cmd /c "echo a & pause"'
let path = 'C:\Users\JanEdmund\tmp\太郎\foo1.bat'
call writefile([cmd], path)
call system(path)

let path = 'C:\Users\JanEdmund\tmp\太郎\foo2.bat'
call writefile([cmd."\r"], path)
call system('"'.path)

let path = 'C:\Users\JanEdmund\tmp\太郎\foo3.bat'
call writefile([cmd], path)
call system('cd '.fnamemodify(path, ':p:h:h').' & ".\太郎\foo3.bat"')
```

watagashi:
@janlazo Simply I think, as long as the batch file always uses CRLF, it will not cause any errors in all Windows environments. 🙂

janlazo:
Ok. Let's go with your fix and use it on fzf's batchfiles as well.

PR #765, opened 2018-06-12
[mod] win32 is enough for check

dohq:
check windows os are no problems with `has('win32') `


junegunn:
For future reference, `:help feature-list` says

```
win32			Win32 version of Vim (MS-Windows 95 and later, 32 or 64 bits)
```

Merging this, thanks.

PR #767, opened 2018-06-13
Added PlugConfig

m42e:
I'd like to keep things separated. This allows to have a config file per plugin, easy accessible and managed. Additionally for lazy loaded plugins sometimes the configuration must also be done after loading. This is included, just add water.

Edit: Fixed some formatting issues.

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,6 +153,7 @@ Reload .vimrc and `:PlugInstall` to install plugins.
 | `PlugStatus`                        | Check the status of plugins                                        |
 | `PlugDiff`                          | Examine changes from the previous update and the pending changes   |
 | `PlugSnapshot[!] [output path]`     | Generate script for restoring the current snapshot of the plugins  |
+| `PlugConfig name`                   | Open or create a config file for a plugin                          |
 
 ### `Plug` options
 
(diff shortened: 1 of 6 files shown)


m42e:
@junegunn Please comment if this seems useful to you or not. Thanks a lot.

junegunn:
Hi, thanks for your interest in the project.

How one organizes the settings for plugins is highly personal. In my case, I prefer having a single, long vimrc file instead of multiple files for each plugin. There's no right or wrong way, it's just a matter of personal preference. The users, especially newcomers to Vim, are better off discovering what's best for them. For that reason, I think vim-plug should not have a strong opinion on that matter and the suggested feature is not a necessary addition to vim-plug.

If I'm not mistaken, the scheme can be implemented without modifying vim-plug but using the information stored in `g:plugs` and `User` autocmd for lazily loaded plugin. Let me know if it's not the case. You might want to check out some examples of additional features that are implemented outside plug.vim in https://github.com/junegunn/vim-plug/wiki/extra. You could probably turn this into a separate, third-party plugin.

m42e:
Thanks for your response.

I got your point. Of course no-one should be tightened down to use a certain method of config management. 

The way of using the User autocmd will work for lazy loaded plugins, not for the other ones.

As there is no heavy interest in getting this into the mainline, I'll close the PR and keep my own fork updated.

m42e:
My apology for digging out this thread. I ported it for me till now, every time I updated. I try now to implement it as vim.plug plugin. But I am curious how I could get a trigger after all the plugins have been loaded. I could not detect an event like that.

m42e:
@junegunn ok, forget the question. I figured it out, here is the plugin: https://github.com/m42e/vim-plug-config maybe you can add a link somewhere if you think it is useful for others, too.

PR #773, opened 2018-06-28
Respect git managed tags

h-east:
Hi Junegunn,

Vim's documentation recommends that package include doc/tags.

:help package-create

```
Run the :helptags command to generate the doc/tags file.  Including this
generated file in the package means that the user can drop the package in his
pack directory and the help command works right away.  Don't forget to re-run
the command after changing the plugin help:
        :helptags path/start/foobar/doc
        :helptags path/opt/fooextra/doc
```

:help help-writing

```
- Make a package with all the files and the tags file available for download.
  Users can drop it in one of the "doc" directories and start use it.
  Report this to Bram, so that he can add a link on www.vim.org.
```

---
In fact, many packages does not contain doc/tags in the repository.
So I think that vim-plug is unconditionally running `:helptags` after installation and update.
I think that should not run `:helptags` in the case of installation or update of that package.
Please respect tags that git managed.

This PR is changed to execute `:helptags` only when either of the following conditions is satisfied.
- tags file is not exist.
- tags file is not managed by git.

Please check it out.

Note:
Repository of package containing doc/tags (or doc/tags-??x)
- https://github.com/vim-jp/vimdoc-ja

--
Best regards,
Hirohito Higashi (h_east)

### DIFF
--- a/plug.vim
+++ b/plug.vim
@@ -600,13 +600,28 @@ function! s:update(force, names)
   call s:update_impl(1, a:force, a:names)
 endfunction
 
+function! s:is_exec_helptags(docdir)
+  execute 'lcd' a:docdir
+  let result = v:false
+  let exts = uniq(sort(map(s:glob(a:docdir, '*.{txt,??x}'), 'v:val[-3:]')))
+  for ext in exts
+    let tagname = 'tags' . (ext == 'txt' ? '' : '-' . ext[:1])
+    if !filereadable(tagname) || empty(s:system('git ls-files ' . tagname))
+      let result = v:true
+      break
+    endif
+  endfor
+  lcd -
+  return result
+endfunc
+
 function! plug#helptags()
   if !exists('g:plugs')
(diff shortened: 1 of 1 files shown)


plug.vim
```diff
@@ -600,13 +600,24 @@ function! s:update(force, names)
   call s:update_impl(1, a:force, a:names)
 endfunction
 
+function! s:is_exec_helptags(docdir)
```
mattn: This function name is un-clear to be what should do.

h-east: How about `s:is_need_exec_helptags` ?

mattn: Since I'm not native speaker, I cannot answer best. But has_helptags(x) or helptags_exists(x) ?
cc @junegunn

h-east: I think `do_need_exec_helptags()` is best.
I'll change it after I came back today or tomorrow. :kissing_smiling_eyes:

Thank you mattn :+1:

PR #791, opened 2018-10-11
Implement ability to override git clone options for a particular plugin

satiani:
This PR is intended to allow users to manually override how git clones a particular repository. An example use case would be when trying to install the `LanguageClient-neovim` plugin. If that plugin is installed like this:

```
Plug 'autozimu/LanguageClient-neovim', {
    \ 'branch': 'next',
    \ 'do': 'bash install.sh',
    \ }
```

It will download ~97MB, even when `g:plug_shallow` is set to 1. The reason is that `plug.vim` still specifies `--no-single-branch` on git versions more than 1.7.10 as can be seen from these lines in `plug.vim`:

```
let s:clone_opt = get(g:, 'plug_shallow', 1) ?
        \ '--depth 1' . (s:git_version_requirement(1, 7, 10) ? ' --no-single-branch' : '') : ''
```

So while we're only interested in the `next` branch specified in the configuration above, it downloads the entire .git history, including some very large binary files that are left over from older commits. This PR makes it possible to do the following:

```
Plug 'autozimu/LanguageClient-neovim', {
    \ 'branch': 'next',
    \ 'do': 'bash install.sh',
    \ 'clone_opt': '--depth 1'
    \ }
```

Which omits the `--no-single-branch` option. In turn that dramatically reduces the size of the repo down to 300KB.

There are some potential issues I can think of with this PR, for example, allowing users to specify arbitrary git clone options may break other VimPlug functionality. For example, without `--no-single-branch`, changing the branch and doing `PlugUpdate` no longer works correctly. This might be an acceptable tradeoff for some users (it is for me), but might not be for the general use case.

Please let me know what you think. If there is anything I can do to improve this PR I'd be happy to do that.

### DIFF
--- a/doc/plug.txt
+++ b/doc/plug.txt
@@ -216,6 +216,7 @@ Reload .vimrc and `:PlugInstall` to install plugins.
   `on`                     | On-demand loading: Commands or <Plug>-mappings
   `for`                    | On-demand loading: File types
   `frozen`                 | Do not update unless explicitly specified
+  `clone_opt`              | Override git cloning options for the plugin
  ------------------------+-----------------------------------------------
 
 
--- a/plug.vim
+++ b/plug.vim
@@ -1337,6 +1337,8 @@ while 1 " Without TCO, Vim stack is bound to explode
   redraw
 
   let has_tag = has_key(spec, 'tag')
+  let has_clone_opt = has_key(spec, 'clone_opt')
+  let clone_opt = has_clone_opt ? spec.clone_opt : s:clone_opt
   if !new
     let [error, _] = s:git_validate(spec, 0)
     if empty(error)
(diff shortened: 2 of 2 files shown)


satiani:
Hmm, not sure why the build failed, I don't see any error messages. Would appreciate some help pointing out the error to me so I can fix it.

junegunn:
> allowing users to specify arbitrary git clone options may break other VimPlug functionality

Yes, I'm reluctant to merge this because of the very reason you mentioned. vim-plug is used by many users with very different levels of expertise in Vim environment, and it's crucial to make it hard to misuse it by keeping it simple and less configurable. So I suggest that you use your own fork if this new feature is important to you. vim-plug has been quite stable for a while and you probably will not see any dramatic change in the near future, so you won't have to worry about merging upstream changes too much.

satiani:
Makes sense, thanks for the response @junegunn

PR #798, opened 2018-10-25
feat(diff): use 'rtp' option to filter diff

### LINKED ISSUE #797: Feature request: filter PlugDiff by 'rtp' option
I manage one of my plugins (https://gn.googlesource.com/gn) with the `rtp` option. The repo receives frequent updates, but the [plugin itself](https://gn.googlesource.com/gn/+log/master/tools/gn/misc/vim) rarely gets modified.

It would be nice to use `v.rtp` to filter out non-plugin commits when I run `:PlugDiff`.

- Type:
    - [ ] Bug
    - [x] Enhancement
    - [x] Feature Request
    - [ ] Question
- OS:
    - [x] All/Other
    - [ ] Linux
    - [ ] OS X
    - [ ] Windows
- Vim:
    - [x] Terminal Vim
    - [x] GVim
    - [x] Neovim

nfischer:
Previously, `:PlugDiff` would show every new commit to a plugin's git
repo. This makes sense for the general case, but makes less sense when a
plugin lives in a subdirectory of the repo (and is configured with the
'rtp' option). This makes it difficult to determine which commits relate
to the plugin and which are unrelated.

This changes `:PlugDiff` to filter out any commits outside of the 'rtp'
folder.

Some consequences:

 * This does not change the `:PlugUpdate` UI. This means `:PlugUpdate`
   may pull down non-plugin commits, display that it has updated the
   plugin, and then `:PlugDiff` will show no updates (since such commits
   fall out of the 'rtp' path).
 * It also means there's no UI to revert non-plugin updates, as they
   don't show up in `:PlugDiff`.

Fixes #797
Test: Manual on Linux machine

### DIFF
--- a/plug.vim
+++ b/plug.vim
@@ -2420,7 +2420,11 @@ function! s:diff()
     call s:append_ul(2, origin ? 'Pending updates:' : 'Last update:')
     for [k, v] in plugs
       let range = origin ? '..origin/'.v.branch : 'HEAD@{1}..'
-      let diff = s:system_chomp('git log --graph --color=never '.join(map(['--pretty=format:%x01%h%x01%d%x01%s%x01%cr', range], 's:shellesc(v:val)')).' -- .', s:rtp(v))
+      let cmd = 'git log --graph --color=never '.join(map(['--pretty=format:%x01%h%x01%d%x01%s%x01%cr', range], 's:shellesc(v:val)'))
+      if has_key(v, 'rtp')
+        let cmd .= ' -- ' . v.rtp
+      endif
+      let diff = s:system_chomp(cmd, v.dir)
       if !empty(diff)
         let ref = has_key(v, 'tag') ? (' (tag: '.v.tag.')') : has_key(v, 'commit') ? (' '.v.commit) : ''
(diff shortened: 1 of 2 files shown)


nfischer:
I see this breaks tests on my local machine, but I can't figure out why it would do so. I could cheat by only applying "` -- .`" if the `rtp` option is supplied (tests will pass), but this makes me wonder if the patch would fail under legitimate circumstances. Any help debugging would be appreciated.

Also, I'm happy to change this to an opt-in feature if the noted consequences are concerning.

nfischer:
Oh, now I see. 2 issues going on with the PR:

 1. It's not correct to unconditionally apply ` -- .`, because a repo may have empty commits (`--allow-empty`). These are used all over tests, which is why some tests broke.
 2. If any test breaks, it appears to halt execution at the first failed assumption. Unfortunately, this often leaves an extra window open (the `:q`-cleanup is at the end of the test). So, each failed test may cause side effects, which caused some other failures related to `winnr('$')` (which has now changed).

junegunn:
Hi, thanks for the patch! I just made a small change to your branch. I'll merge this once the test passes.

nfischer:
Thanks, @junegunn (and good catch)

PR #811, opened 2018-12-03
Correct installation command in README

liuxh0:
In README, the installation command for Neovim and Unix uses another path than in Wiki > Tutorial > Setting up.

The path in README doesn't work for me on a Mac, while the one in Tutorial works.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ $uri = 'https://raw.githubusercontent.com/junegunn/vim-plug/master/plug.vim'
 ###### Unix
 
 ```sh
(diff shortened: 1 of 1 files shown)


junegunn:
I believe the path shown in the README page is the correct one, as it was suggested by the maintainer of Neovim. Please see https://github.com/junegunn/vim-plug/issues/559. I just updated the tutorial page accordingly.

liuxh0:
I checked it again and you are right!
Sorry my bad.

junegunn:
No problem. Thanks.

PR #816, opened 2019-01-03
Tuning4neovim

stappersg:
Hi,

Seeing _vim_ specific instructions in `plug.vim` did confuse me.

This pull request is an attempt to get less _vim noise_ for neovim users.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,14 +60,14 @@ $uri = 'https://raw.githubusercontent.com/junegunn/vim-plug/master/plug.vim'
 
 ```sh
 curl -fLo ~/.local/share/nvim/site/autoload/plug.vim --create-dirs \
-    https://raw.githubusercontent.com/junegunn/vim-plug/master/plug.vim
+    https://raw.githubusercontent.com/junegunn/vim-plug/master/plug.neovim
 ```
 
 ###### Windows (PowerShell)
 
 ```powershell
 md ~\AppData\Local\nvim\autoload
-$uri = 'https://raw.githubusercontent.com/junegunn/vim-plug/master/plug.vim'
+$uri = 'https://raw.githubusercontent.com/junegunn/vim-plug/master/plug.neovim'
 (New-Object Net.WebClient).DownloadFile(
   $uri,
   $ExecutionContext.SessionState.Path.GetUnresolvedProviderPathFromPSPath(
--- a/doc/plug.txt
+++ b/doc/plug.txt
@@ -94,13 +94,13 @@ Neovim~
 >> Unix~
 >
     curl -fLo ~/.local/share/nvim/site/autoload/plug.vim --create-dirs \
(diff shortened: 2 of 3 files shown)


pmav99:
By the way instead of having to maintain two sets of `*.vim` files, it might make sense just to improve the example in the documentation. E.g.

``` vim
" Bootstrap Plug {
  " Set the appropriate paths for Plug
  " PLUG_AUTOLOAD_PATH is a script variable because we only need to bootstrap
  " Plug on first run.
  " PLUG_DATA_DIR is a global because we might want to use it in .gvimrc etc
  if has('nvim')
    let s:PLUG_AUTOLOAD_PATH = stdpath('data') . '/site/autoload/plug.vim'
    let g:PLUG_DATA_DIR = stdpath('data') . '/plugged'
  else
    let s:PLUG_AUTOLOAD_PATH = $HOME . '/.vim/autoload/plug.vim'
    let g:PLUG_DATA_DIR = $HOME . '/.vim/plugged'
  endif

  " Install plug
  if !filereadable(s:PLUG_AUTOLOAD_PATH)
    silent execute '!curl -fLo ' . s:PLUG_AUTOLOAD_PATH . ' --create-dirs "https://raw.githubusercontent.com/junegunn/vim-plug/master/plug.vim"'
    autocmd VimEnter * PlugInstall --sync | source $MYVIMRC
  endif
  unlet s:PLUG_AUTOLOAD_PATH
" Bootstrap Plug }

" Plugins {
  call plug#begin(g:PLUG_DATA_DIR)
    Plug 'itchyny/lightline.vim'                                                    " Status Line
    Plug 'scrooloose/nerdcommenter'                                                 " Comments
    Plug 'mbbill/undotree'                                                          " Undo
    Plug 'editorconfig/editorconfig-vim'
    Plug 'tpope/vim-eunuch'
    Plug 'tpope/vim-repeat'
    Plug 'tpope/vim-surround'
    if filereadable("/usr/bin/fzf")                                                 " Fuzzy
      Plug 'junegunn/fzf.vim'
    endif
  call plug#end()

  " if we want to define more plugins in a different file we can do it like this
  call plug#begin(g:PLUG_DATA_DIR)
    Plug 'morhetz/gruvbox'
  call plug#end()
" Plugins }

```

junegunn:
Sorry, I do not like this approach.

pmav99:
@junegunn How about the bootstrap code in my previous message? Would you accept a PR with that?

junegunn:
I can update [the wiki page](https://github.com/junegunn/vim-plug/wiki/tips#automatic-installation), but I don't want to present the code upfront in the README page, because some people will recognize it as necessary boilerplate code when it's just optional. Also, both Vim and Neovim users on Windows platform will complain that the code doesn't run on their system, unless we provide code for 4 different sets of paths. I'd like to keep README page as short as possible.

pmav99:
sure, no problem

PR #853, opened 2019-06-28
Change PlugClean description

morenoh149:
this description is more clear about uninstalling plugins


junegunn:
Makes sense, can you also update https://github.com/junegunn/vim-plug/blob/master/doc/plug.txt#L197?

junegunn:
Merged, thanks.

Issue #815, opened 2018-12-25
When the &shell is the Bash of Cygwin in MS-Windows system.


mikelue:
I am using MS-Windows system with **the Bash of Cygwin**, which is my `&shell` of vim.
I have tried to make a patch for supporting default behavior of UNIX-like.
[plug.vim.patch.zip](https://github.com/junegunn/vim-plug/files/2708253/plug.vim.patch.zip)

```diff
-let s:is_win = has('win32')
+let s:is_win = has('win32') && &shell !~# 'sh'
```

Additionaly, I have set `$SHELL` environment variable to the Bash of Cygwin in windows system.
Thanks a lot.

------------------------------

```
VIM - Vi IMproved 8.1 (2018 May 18, compiled Nov 30 2018 23:02:27)
MS-Windows 64-bit GUI version with OLE support
Included patches: 1-553
Compiled by appveyor@APPVYR-WIN
Huge version with GUI.  Features included (+) or not (-):
+acl                +browse             +cmdline_compl      +cursorbind         -ebcdic             +find_in_path       +insert_expand      +linebreak          +modify_fname       +num64              +printer            +ruby/dyn           +syntax             -tgetent            +vartabs            -vtp                +xpm_w32            
+arabic             ++builtin_terms     +cmdline_hist       +cursorshape        +emacs_tags         +float              +job                +lispindent         +mouse              +ole                +profile            +scrollbind         +tag_binary         -termresponse       +vertsplit          +wildignore         -xterm_save         
+autocmd            +byte_offset        +cmdline_info       +dialog_con_gui     +eval               +folding            +jumplist           +listcmds           +mouseshape         +packages           +python/dyn         +signs              +tag_old_static     +textobjects        +virtualedit        +wildmenu           
+autochdir          +channel            +comments           +diff               +ex_extra           -footer             +keymap             +localmap           +multi_byte_ime/dyn +path_extra         +python3/dyn        +smartindent        -tag_any_white      +timers             +visual             +windows            
+autoservername     +cindent            +conceal            +digraphs           +extra_search       +gettext/dyn        +lambda             +lua/dyn            +multi_lang         +perl/dyn           +quickfix           +startuptime        +tcl/dyn            +title              +visualextra        +writebackup        
[...]

PR #856, opened 2019-07-12
Don't override shell on Windows

janlazo:
Assume that the user set the shell options correctly before running vim-plug so that the user can use bash or powershell as their shell.

If this works, port the fix to fzf's Vim plugin to resolve related issues. If there's time, avoid the batchfile on Neovim or Vim jobs before porting to fzf.

Close https://github.com/junegunn/vim-plug/issues/815

### DIFF
--- a/plug.vim
+++ b/plug.vim
@@ -794,9 +794,7 @@ endfunction
 
 function! s:chsh(swap)
   let prev = [&shell, &shellcmdflag, &shellredir]
-  if s:is_win
-    set shell=cmd.exe shellcmdflag=/c shellredir=>%s\ 2>&1
-  elseif a:swap
+  if !s:is_win && a:swap
     set shell=sh shellredir=>%s\ 2>&1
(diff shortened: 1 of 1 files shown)


plug.vim
```diff
@@ -811,7 +809,7 @@ function! s:bang(cmd, ...)
     if s:is_win
       let batchfile = tempname().'.bat'
       call writefile(["@echo off\r", cmd . "\r"], batchfile)
-      let cmd = s:shellesc(batchfile)
+      let cmd = s:shellesc(expand(batchfile))
```
janlazo: This should be enough on Cygwin. If not, try `fnamemodify(batchfile, ':p')` if `has('win32unix')` is true. If not enough, try [cygpath](http://cygwin-lite.sourceforge.net/html/cygpath.html) or similar utilities.

junegunn:
Merged, thanks!

PR #471, opened 2016-04-13
Don't pass "-f .git/config" to git config

qstrahl:
- Fixes #470

PR #858, opened 2019-07-14
remove `-f .git/config` from `git config`

tianyuanhao:
`-f` is redundant and results in an error when `GIT_DIR` is not `.git`.


tianyuanhao:
I found that I have done the similar thing as #471.

`GIT_CONFIG` only affects `git config` as documented [here](https://github.com/git/git/blob/master/Documentation/RelNotes/1.6.0.txt). I thought there should be few people use it. Why don't we just unset it, so that `:PlugUpdate` works properly whether plugins are installed by `vim-plug` or `git submodule`.

plug.vim
```diff
@@ -1625,7 +1625,7 @@ class Plugin(object):
     self.write(Action.DONE, self.name, result[-1:])
 
   def repo_uri(self):
-    cmd = 'git rev-parse --abbrev-ref HEAD 2>&1 && git config -f .git/config remote.origin.url'
+    cmd = 'git rev-parse --abbrev-ref HEAD 2>&1 && unset GIT_CONFIG && git config remote.origin.url'
```
erydo: Should the unset be done before the first `git rev-parse`, I wonder? Could an errant `GIT_CONFIG` somehow change how `--abbrev-ref` determines the branch name?

erydo: Re my comment, if that's the case, note that that would be an _existing_ bug rather than a problem introduced by this PR. I think this PR could be merged as is.

plug.vim
```diff
@@ -2054,7 +2054,7 @@ endfunction
 function! s:git_validate(spec, check_branch)
   let err = ''
   if isdirectory(a:spec.dir)
-    let result = s:lines(s:system('git rev-parse --abbrev-ref HEAD 2>&1 && git config -f .git/config remote.origin.url', a:spec.dir))
```
janlazo: Is this meant to work on all OS?

tianyuanhao: Well, actually only tested on Linux.

Curry2323:
Yygtt

Curry2323:
Conflicting files plug.vim

Issue #176, opened 2015-02-18
Travis Spends Most Time Building Vim


starcraftman:
From recent build, the second one is the nox one that just installs a version of vim. Total is usually around 15 mins now. If we didn't build vm, would probably be like 3/4.
 93.1   4 min 34 sec    3 minutes ago   1.8.7   linux
93.2    1 min           7 minutes ago   1.9.2   linux
93.3    5 min 27 sec    2 minutes ago   1.9.3   linux
93.4    4 min 52 sec    3 minutes ago   2.0.0   linux

Build time for all cases except nox are lengthy not because of our project but because the initial hg clone is really long and the subsequent build is quite long too.

All we need to do is prebuild a compatible version of vim (with dynamic python/ruby) I guess. I mean it isn't like we absolutely always need to test the immediate latest vim tag. I might investigate this in a bit on a separate branch. 
- Zip/Extract onto machine
- Have some Ubuntu PPA with our own prebuild deb

junegunn:
I know. It would be nice if we could cut down the time, but more important I think is that we make sure vim-plug works with various versions of Vim, which is the point that the current set of Travis CI builds is missing. I frequently have to work with the ancient RHEL 5 servers with Vim 7.0 at work, and vim-plug seems to work fine there. I wish such old versions of Vim are also tested on Travis CI.

starcraftman:
@junegunn Right, too bad travis doesn't seem to be looking at that so far. Only thing I see is OSX on horizon. Currently might have ruby support, but python non functional > https://github.com/travis-ci/travis-ci/issues/2320

starcraftman:
@junegunn Putting this on hold while I look at other stuff like recent bug.

I tried to create a deb & extract it from travis and that worked. Seem to have a small problem with vim runtimedir when reinstalled. Small error to solve. 

Alternatively, noticed that travis has arbitrary directory caching. http://docs.travis-ci.com/user/caching/#Fetching-and-storing-caches

If we made our vim builds sit in some dir under home like $HOME/cache, we could then maybe cache them between builds & conditionally build if cache was empty.

junegunn:
@starcraftman The page you linked says:

> The features described here are currently only available for private repositories on travis-ci.com and our new container-based infrastructure.

So it doesn't seem like an option. Anyway, thanks for looking into this.

starcraftman:
[...]

PR #867, opened 2019-08-10
Travis: add ubuntu xenial and bionic

janlazo:
Ubuntu Trusty reached EOL since April 2019.

Close https://github.com/junegunn/vim-plug/issues/176

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,54 +1,51 @@
-language: ruby
-dist: trusty
-sudo: false
+language: minimal
 env:
   global:
     - DEPS=$HOME/deps
     - PATH=$DEPS/bin:$PATH
 matrix:
   include:
-    - env: ENV=vim74
-      rvm: 1.8.7
-      addons: { apt: { packages: [vim-nox] } }
-    - env: ENV=python
-      rvm: 1.8.7
-      addons: { apt: { packages: [python2.7-dev] } }
-    - env: ENV=python3
-      rvm: 1.8.7
-      addons: { apt: { packages: [python3-dev] } }
-    - env: ENV=ruby18
-      rvm: 1.8.7
-    - env: ENV=ruby20
-      rvm: 2.0.0
-    - env: ENV=neovim
-    - env: ENV=vim8
+    - env: ENV=vim80-bionic
+      dist: bionic
+    - env: ENV=vim-nightly
+      dist: trusty
+    - env: ENV=neovim-stable
+      dist: xenial
+      addons: {apt: {packages: [neovim], sources: [{sourceline: 'ppa:neovim-ppa/stable'}]}}
+    - env: ENV=neovim-nightly
+      dist: xenial
+      addons: {apt: {packages: [neovim], sources: [{sourceline: 'ppa:neovim-ppa/unstable'}]}}
+    - env: ENV=vim74-trusty-python
+      dist: trusty
+    - env: ENV=vim74-xenial-python3
+      dist: xenial
+    - env: ENV=vim74-trusty-ruby
+      dist: trusty
+      addons: {apt: {packages: [vim-nox]}}
+    - env: ENV=vim74-xenial-ruby
+      dist: xenial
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -8,18 +8,18 @@ env:
 matrix:
   include:
     - env: ENV=vim74
-      rvm: 1.8.7
+      rvm: 2.1.5
```
janlazo: 1.8.7 didn't work in Xenial. Is there a usecase to continue testing ruby versions before 2.1?

junegunn: Probably not, but how about moving `dist` into `matrix` so that both distro versions are tested with their accompanying Ruby versions?


```yaml
    - env: ENV=vim74
      dist: trusty
      rvm: 1.8.7
      addons: { apt: { packages: [vim-nox] } }
    - env: ENV=vim74
      dist: xenial
      rvm: 2.1.5
      addons: { apt: { packages: [vim-nox] } }
...
```

Ref: https://docs.travis-ci.com/user/multi-os/

.travis.yml
```diff
@@ -1,53 +1,94 @@
...
+      dist: xenial
+      addons: {apt: {packages: [neovim], sources: [{sourceline: 'ppa:neovim-ppa/unstable'}]}}
+    - env: ENV=neovim-bionic
+      dist: bionic
+      addons: {apt: {packages: [neovim], sources: [{sourceline: 'ppa:neovim-ppa/unstable'}]}}
     - env: ENV=vim8
+      dist: trusty
+    - env: ENV=vim8
+      dist: xenial
+    - env: ENV=vim8
```
janlazo: Is this needed now that Ubuntu bionic has Vim 8?

.travis.yml
```diff
@@ -1,53 +1,94 @@
...
+    vim74-* | vim80-*)
+      mkdir -p ${DEPS}/bin
+      ln -s /usr/bin/vim.nox ${DEPS}/bin/vim
+      return
+      ;;
+    python-*)
       C_OPTS+=--enable-pythoninterp
       ;;
-    python3)
+    python3-*)
```
janlazo: Is this suppose to test on Vim nightly (8.1) or the Vim version that Ubuntu has (7.4, 8.0)?

janlazo:
is `$ENV` hack to detect vim 8 still needed in test.vader? I don't know what it's suppose to do. Isn't `v:version` or `has('patch-8')` enough for version detection?

.travis.yml
junegunn: The latter. Ruby/Python installers are no longer used on the latest versions of Vim.

junegunn:
I guess the purpose of the ENV hack was to test Ruby or Python installer even on Vim 8.

https://github.com/junegunn/vim-plug/blob/226d6abeb2f02bcd8f5b11288543259fa2e02962/test/test.vader#L21-L23

That was when the job feature of Vim 8 was new, unstable, and not widely available. Three years later, it doesn't make much sense to test Ruby/Python installers with Vim 8 anymore.

So basically we need to test:
1. with Vim 8 or later (nightly)
1. with Neovim
1. with Vim 7.x for Ruby installer
1. with Vim 7.x for Python installer
1. with Vim 7.x for Python3 installer

.travis.yml
```diff
@@ -1,55 +1,61 @@
...
+      ln -s /usr/bin/nvim ${DEPS}/bin/vim
+      export VADER_OUTPUT_FILE=/dev/stderr
+      return
+      ;;
+    vim74-* | vim80-*)
+      mkdir -p ${DEPS}/bin
+      ln -s /usr/bin/vim.nox ${DEPS}/bin/vim
+      return
+      ;;
+    python-*)
```
junegunn: We can now remove `python-*`, `python3-*`, and `ruby-*`. Right?

janlazo: Yes. I left it there in case you want to test on a custom Vim 7.x.

junegunn:
Thanks!

PR #860, opened 2019-07-23
Fix Windows support for Unix shells and powershell

janlazo:
Fix #815 
Fix (partial) #690
Fix https://github.com/junegunn/vim-plug/issues/831
Fix https://github.com/junegunn/vim-plug/issues/832 (issue happens on Windows for custom shell settings without this PR)

@mikelue Can you test if it works?

TODO
- [X] powershell
   - require `set noshellslash`, same as cmd.exe
  - fails on `do` hook after installing the repo (tested with `do` hook to install the fzf binary)  
  - `shellredir` must be set to `| Out-File -Encoding utf8` 
- [ ] figure out how to port this to fzf Vim plugin (`fzf#shellescape` uses a different shell default)
- [x] investigate if Vim escapes executable arguments in `job_start([executable, arg1, arg2, ...])` and if it's still broken on cmd.exe, similar to `term_start` behavior.
   - Still broken in 8.1.1794 so the hack in vim-plug should be ported to fzf

PR #870, opened 2019-08-17
Travis: add macOS 10.14

janlazo:
I don't have a mac to verify that my changes don't introduce any regressions.

Ref: https://github.com/junegunn/vim-plug/pull/860#issuecomment-521964314

`osx_image` (https://docs.travis-ci.com/user/reference/osx/#macos-version) should be recent enough to avoid outdated homebrew packages because Travis may take 5-10 mins updating homebrew.


junegunn:
Thanks!

PR #875, opened 2019-09-07
Include function to check if plugin is loaded

### LINKED ISSUE #849: hello how to know whether a plug has loaded?
Explain the problem here ...

```
hello, how can i know whether a plugin has loaded?
```

- Type:
    - [ ] Bug
    - [ ] Enhancement
    - [ ] Feature Request
    - [X] Question
- OS:
    - [ ] All/Other
    - [X] Linux
    - [ ] OS X
    - [ ] Windows
- Vim:
    - [ ] Terminal Vim
    - [ ] GVim
    - [X] Neovim

kyounger:
Simple public function that checks whether a plugin is loaded.

Fixes #849 

This allows me to do a snippet lower in my init.vim such as:

```
if plug#is_loaded('NeoSolarized')
    colorscheme NeoSolarized
endif
```

(I'm more than open to modifying this if needed, but thought a little code was a good start to the PR process.)

### DIFF
--- a/plug.vim
+++ b/plug.vim
@@ -2260,6 +2260,10 @@ function! s:upgrade_specs()
   endfor
 endfunction
 
+function! plug#is_loaded(name)
+    return has_key(s:loaded, a:name)
+endfunction
+
 function! s:status()
   call s:prepare()
   call append(0, 'Checking plugins')


junegunn:
Thanks for the suggestion, but I don't think it's really needed. vim-plug tries to be as minimal as possible, and we don't add a feature when there are easy workarounds.

1. In that case, you can simply ignore the error message by prepending `silent!`.

```vim
silent! colorscheme NeoSolarized
```

That's how I've been doing it in my vimrc:

https://github.com/junegunn/dotfiles/blob/3a0e6778c91073d19c37dccdb34a8213811abf03/vimrc#L274-L279


2. In other cases, you can check if a plugin is loaded by examining the `&rtp` like so:

```vim
if stridx(&rtp, g:plugs['NeoSolarized'].dir)
  ...
endif
```

kyounger:
No worries, thanks for the feedback. I'll try your second snippet!

kyounger:
Ah, for the benefit of others coming here, you need to check that the key exists in g:plugs first :)

`has_key(g:plugs, a:plugin_name) && stridx(&rtp, g:plugs[a:plugin_name].dir)`

PR #886, opened 2019-09-28
output of chcp was not parsed correctly

gh4w:
On Windows, when wrapping a batch command with the function s:wrap_cmds(),
when calling 'chcp' to get the current code page, the code assumed that
the output was in the format "active code page: XXX" (where XXX is the
code page), whereas the actual output is localized (for instance, in
French, the output would be: "page de code active: XXX").
The parsing of the output relied on that, and this failed for a
message different from "active code page" (i.e., English).

This patch changes the parsing to split the output of chcp on the colon
instead of spaces. Assuming that the output is always in the format
"localized message: XXX", regardless of the locale, hopefully, this is
a bit more robust.

### DIFF
--- a/plug.vim
+++ b/plug.vim
@@ -353,9 +353,9 @@ if s:is_win
 
   " Copied from fzf
   function! s:wrap_cmds(cmds)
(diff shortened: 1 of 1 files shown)


janlazo:
Thanks @gh4w for the fix and using `deleayexpansion`.

Issue #884, opened 2019-09-21
chcp parsing is wrong for German locale


wngrtn:
PlugUpdate fails to update my plugins, but it doesn't show any error messages.

How do I diagnose this or make vim-plug give me more verbose error messages?

Here is the output I am getting:

```
Updated. Elapsed time: 4.042648 sec.
[xxxxxxxxxxxxxxxxxxxxxxxxxxxx]

- Finishing ... Done!
x vim-polyglot:
    OK
x fzf:
    OK
x vim-devicons:
    OK
x tagalong.vim:
    OK
```

(Concatenated, but it continues like this for all plugins.)

------------------------------

```
NVIM v0.4.2
Build type: RelWithDebInfo
LuaJIT 2.0.5
Compilation: C:/Program Files (x86)/Microsoft Visual Studio/2017/Community/VC/Tools/MSVC/14.16.27023/bin/Hostx86/x64/cl.exe /DWIN32 /D_WINDOWS /W3 /MD /Zi /O2 /Ob1 /DNDEBUG -DMIN_LOG_LEVEL=3 /W3 -D_CRT_SECURE_NO_WARNINGS -D_CRT_NONSTDC_NO_DEPRECATE -DWIN32 -D_WIN32_WINNT=0x0600 -DINCLUDE_GENERATED_DECLARATIONS -DNVIM_MSGPACK_HAS_FLOAT32 -DNVIM_UNIBI_HAS_VAR_FROM -IC:/projects/neovim/build/config -IC:/projects/neovim/src -IC:/projects/nvim-deps/usr/include -IC:/projects/neovim/build/src/nvim/auto -IC:/projects/neovim/build/include

übersetzt    von appveyor@APPVYR-WIN

Features: -acl +iconv +tui
See ":help feature-compile"

          System-vimrc-Datei: "$VIM\sysinit.vim"
     Voreinstellung für $VIM: "   C:/Program Files/nvim/share/nvim"

Run :checkhealth for more info
```

- Type:
    - [x] Bug
    - [ ] Enhancement
    - [ ] Feature Request
    - [x] Question
- OS:
    - [ ] All/Other
    - [ ] Linux
    - [ ] OS X
    - [x] Windows
- Vim:
    - [ ] Terminal Vim
    - [ ] GVim
    - [x] Neovim

flotisable:
I have a similar problem but it shows "PlugClean required" for each plugin.
And this happens only when I set the shell option to be **powershell**,
[...]

PR #891, opened 2019-10-13
Use chcp only if sed is in PATH

janlazo:
Fix https://github.com/junegunn/vim-plug/issues/884

`sed` is now required to parse `chcp` output. Vim can be used instead via separate child `system` call but doing it per `s:system` makes Windows slower. `chcp` parsing is fragile because of the system locale. There's no convenient way to parse out the codepage value without regex just by relying on cmd.exe builtins and default binaries in `PATH`.
If this still doesn't work, users will have to get by with ascii output only (no readable unicode symbols from console) because this is a OS problem. This applies to fzf as well because the chcp code orignated from fzf Vim plugin was meant for UTF-8 characters.

Powershell builtins could be used as an alternative to sed but using powershell on non-powershell `&shell` slows down `s:system`.

### DIFF
--- a/plug.vim
+++ b/plug.vim
@@ -353,14 +353,16 @@ if s:is_win
 
   " Copied from fzf
   function! s:wrap_cmds(cmds)
+    let use_chcp = executable('sed')
     return map([
       \ '@echo off',
-      \ 'setlocal enabledelayedexpansion',
-      \ 'for /f "tokens=*" %%a in (''chcp'') do for %%b in (%%a) do set origchcp=%%b',
-      \ 'chcp 65001 > nul'
-    \ ]
+      \ 'setlocal enabledelayedexpansion']
+    \ + (use_chcp ? [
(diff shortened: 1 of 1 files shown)


plug.vim
```diff
@@ -353,14 +353,16 @@ if s:is_win
...
+    let use_chcp = executable('sed')
     return map([
       \ '@echo off',
-      \ 'setlocal enabledelayedexpansion',
-      \ 'for /f "tokens=*" %%a in (''chcp'') do for %%b in (%%a) do set origchcp=%%b',
-      \ 'chcp 65001 > nul'
-    \ ]
+      \ 'setlocal enabledelayedexpansion']
+    \ + (use_chcp ? [
+      \ 'for /f "usebackq" %%a in (`chcp ^| sed "s/[^0-9]//gp"`) do set origchcp=%%a',
```
janlazo: This assumes that the first numeric token is the codepage.

Issue #908, opened 2019-11-16
ERROR "invalid uri ^" when runing PlugStatus


CodingCloud9527:
Explain the problem here ...
When i running PlugStatus, I always get a error:
Finished. 1 error(s).
[=]

x auto-pairs:
    Invalid URI: ^A
    Expected:    https://git::@github.com/jiangmiao/auto-pairs.git
    PlugClean required.

and i see the variable `result` in https://github.com/junegunn/vim-plug/blob/68fef9c2fd9d4a21b500cc2249b6711a71c6fb9f/plug.vim#L2075 is Irregular letters and symbols , just like "^DA^GHF^D^DGA........."
------------------------------

```
VIM - Vi IMproved 8.1 (2018 May 18, compiled Nov 15 2019 23:02:52)
MS-Windows 64-bit console version
包含补丁: 1-2301
```

- Type:
    - [x] Bug
    - [ ] Enhancement
    - [ ] Feature Request
    - [ ] Question
- OS:
    - [ ] All/Other
    - [ ] Linux
    - [ ] OS X
    - [x] Windows
- Vim:
    - [x] Terminal Vim
    - [ ] GVim
    - [ ] Neovim

janlazo:
`git --version` output?
Is `sed` executable in `PATH`?

CodingCloud9527:
hrer is `git --version` output:
`git version 2.24.0.windows.2`
I retry after put `sed` executable in `PATH`,The problem still appears


> `git --version` output?
> Is `sed` executable in `PATH`?

janlazo:
Do you have `set encoding=utf-8' for your console vimrc config and, if you have gvim, does it happen with the same (g)vimrc?

CodingCloud9527:
I have `set encoding=utf-8`, and my gvim has the same problem

mattn:
I god same. I figured out the cause of problem. If using sed provided from msys2 or cygwin, it result from `chcp ^| sed "s/[^0-9]//gp"` is not expected string.

mattn:
This is a patch to fix this.
```diff
diff --git a/plug.vim b/plug.vim
index 02fac8d..b945417 100644
--- a/plug.vim
+++ b/plug.vim
@@ -353,16 +353,11 @@ if s:is_win
 
   " Copied from fzf
   function! s:wrap_cmds(cmds)
-    let use_chcp = executable('sed')
-    return map([
-      \ '@echo off',
[...]

PR #913, opened 2019-11-29
Don't use chcp.

janlazo:
User should fix their terminal font to display unicode characters.
Terminal programs can only use Wide String APIs.
For Vim, this requires +multi_byte feature ~~and `set encoding=utf-8`
in the user's vimrc.~~ `iconv()` and `libcallnr` to get the current codepage (can this be cached?).
Neovim always defaults to `set encoding=utf-8`.

~~vim-plug runs commands in a batchfile with utf-8 characters.
On terminal Vim, run the batchfile in a separate minimized console
to not break the terminal console.~~

https://dev.to/mattn/please-stop-hack-chcp-65001-27db

cc @mattn @junegunn 

Close https://github.com/junegunn/vim-plug/issues/908

### DIFF
--- a/plug.vim
+++ b/plug.vim
@@ -182,6 +182,11 @@ function! s:define_commands()
   \ && (&shell =~# 'cmd\.exe' || &shell =~# 'powershell\.exe')
     return s:err('vim-plug does not support shell, ' . &shell . ', when shellslash is set.')
   endif
+  if !has('nvim')
+    \ && (has('win32') || has('win32unix'))
+    \ && (!has('multi_byte') || !has('iconv'))
+    return s:err('Vim needs +iconv, +multi_byte features on Windows to run shell commands.')
+  endif
   command! -nargs=* -bar -bang -complete=customlist,s:names PlugInstall call s:install(<bang>0, [<f-args>])
   command! -nargs=* -bar -bang -complete=customlist,s:names PlugUpdate  call s:update(<bang>0, [<f-args>])
   command! -nargs=0 -bar -bang PlugClean call s:clean(<bang>0)
@@ -395,18 +400,14 @@ if s:is_win
   endfunction
 
   " Copied from fzf
+  let s:codepage = libcallnr('kernel32.dll', 'GetACP', 0)
   function! s:wrap_cmds(cmds)
-    let use_chcp = executable('sed')
     return map([
       \ '@echo off',
       \ 'setlocal enabledelayedexpansion']
-    \ + (use_chcp ? [
-      \ 'for /f "usebackq" %%a in (`chcp ^| sed "s/[^0-9]//gp"`) do set origchcp=%%a',
-      \ 'chcp 65001 > nul'] : [])
     \ + (type(a:cmds) == type([]) ? a:cmds : [a:cmds])
-    \ + (use_chcp ? ['chcp !origchcp! > nul'] : [])
     \ + ['endlocal'],
-    \ 'v:val."\r"')
+    \ printf('iconv(v:val."\r", "%s", "cp%d")', &encoding, s:codepage))
   endfunction
 
   function! s:batchfile(cmd)


plug.vim
```diff
@@ -353,25 +358,22 @@ if s:is_win
...
     \ + ['endlocal'],
     \ 'v:val."\r"')
   endfunction
 
   function! s:batchfile(cmd)
     let batchfile = tempname().'.bat'
     call writefile(s:wrap_cmds(a:cmd), batchfile)
-    let cmd = plug#shellescape(batchfile, {'shell': &shell, 'script': 1})
-    if &shell =~# 'powershell\.exe$'
+    let cmd = plug#shellescape(batchfile, {'shell': &shell, 'script': 0})
```
janlazo: It should have been `0` to begin with because the batchfile will run on the shell, not inside another batchfile.

plug.vim
```diff
@@ -2005,9 +2007,9 @@ function! plug#shellescape(arg, ...)
   let opts = a:0 > 0 && type(a:1) == s:TYPE.dict ? a:1 : {}
   let shell = get(opts, 'shell', s:is_win ? 'cmd.exe' : 'sh')
   let script = get(opts, 'script', 1)
-  if shell =~# 'cmd\.exe$'
+  if shell =~# 'cmd\.exe'
     return s:shellesc_cmd(a:arg, script)
-  elseif shell =~# 'powershell\.exe$' || shell =~# 'pwsh$'
+  elseif shell =~# 'powershell\.exe' || shell =~# 'pwsh$'
```
janlazo: `$`  was a premature optimization because 8.0.1455 and relatex 8.1.x patches were not considered.

plug.vim
```diff
@@ -139,6 +139,11 @@ function! s:define_commands()
   if !executable('git')
     return s:err('`git` executable not found. Most commands will not be available. To suppress this message, prepend `silent!` to `call plug#begin(...)`.')
   endif
+  if !has('nvim')
+    \ && (has('win32') || has('win32unix'))
+    \ && (!has('multi-byte') || &encoding !=# 'utf-8')
+    return s:err('Only Vim with +multi_byte feature is supported. Add `set encoding=utf-8` in your vimrc before running `plug#begin()`.')
```
janlazo: @mattn Is this a reasonable restriction to guarantee that the batchfile has `utf-8` encoding?

mattn: Why do you check support of encoding=utf-8 ?

plug.vim
```diff
@@ -353,25 +358,22 @@ if s:is_win
...
   function! s:batchfile(cmd)
     let batchfile = tempname().'.bat'
     call writefile(s:wrap_cmds(a:cmd), batchfile)
-    let cmd = plug#shellescape(batchfile, {'shell': &shell, 'script': 1})
-    if &shell =~# 'powershell\.exe$'
+    let cmd = plug#shellescape(batchfile, {'shell': &shell, 'script': 0})
+    if &shell =~# 'powershell\.exe'
       let cmd = '& ' . cmd
+    elseif !has('nvim') && !has('gui_running') && &shell =~# 'cmd\.exe'
+      let cmd = 'start /wait /min ' . cmd
```
mattn: Well, sorry my explaining was too short. I mean that vim-plug must not use 'chcp 932' or 'start /wait' for this iseue here.

mattn:
This is my suggestion.

https://gist.github.com/mattn/9182880f9080d728338bfea9ef1edccd

plug.vim
```diff
@@ -373,7 +373,7 @@ if s:is_win
     if &shell =~# 'powershell\.exe'
       let cmd = '& ' . cmd
     elseif !has('nvim') && !has('gui_running') && &shell =~# 'cmd\.exe'
-      let cmd = 'start /wait /min ' . cmd
+      let cmd = 'start ' . plug#shellescape('fzf', {'script': 0}) . ' /min ' . cmd
```
mattn: I think "start" is not required.

janlazo:
Batchfile is required on Vim to support multiple shells while using cmd.exe syntax and not changing the user's shell settings, which depend on the editor and OS. I will reconsider if https://github.com/junegunn/vim-plug/pull/594 is finished before removing the batchfile code because Vim's `job_start` does not use the user's shell.

plug.vim
janlazo: I want to guarantee that `writefile` works with unicode and batchfile has unicode encoding. I don't know enough about the Vim internals to know if this is needed or is insufficent so I'm asking. Maybe `writefile` with `b` flag is enough?

mattn: The batch file must be written in code page NOT unicode. So you can use iconv().

mattn:
```
let cp = libcallnr('kernel32.dll', 'GetACP', 0)
let text = iconv(text, &encoding, printf('cp%d', cp))
```

janlazo: iconv  replaces unicode characters with `?`. I'd rather use powershell if the batchfile cannot be executed with unicode characters and non-unicode codepage.

mattn: Why you have to handle unicode characters?

plug.vim
```diff
@@ -139,6 +139,11 @@ function! s:define_commands()
   if !executable('git')
     return s:err('`git` executable not found. Most commands will not be available. To suppress this message, prepend `silent!` to `call plug#begin(...)`.')
   endif
+  if !has('nvim')
+    \ && (has('win32') || has('win32unix'))
+    \ && (!has('multi-byte') || !has('iconv'))
```
mattn: s/multi-byte/multi_byte/

mattn:
LGTM

plug.vim
```diff
@@ -352,18 +357,14 @@ if s:is_win
   endfunction
 
   " Copied from fzf
```
junegunn: @janlazo Should I backport this to fzf?

janlazo: Sure, after it's merged. I haven't tested the code myself to verify that it's enough.

janlazo:
Works with my config on a Windows VM. Tested with cmd.exe, powershell.exe, and bash.exe.

janlazo:
@mattn Thanks for the help.

janlazo:
@junegunn https://github.com/junegunn/fzf/commit/f68017d21e78b23fbf648d579edb6500ae7a6de8, port of this PR, breaks filtering for `Commands`. Back to square one :disappointed:

PR #916, opened 2019-12-05
Detect shellslash on Windows

janlazo:
Windows users may have `set shellslash` in their vimrc without changing their shell.

https://www.reddit.com/r/neovim/comments/e1vl6o/plugupdate_not_working/

~~Actual fix is to port https://github.com/junegunn/fzf/commit/7a11a06cbd03353e7c0c7ddc25cd6e9fe58cef6d#diff-24d9d57e35df5afce7890f2400917919R29-R76 from fzf~~ Done. Works on Vim. Neovim breaks if `shellslash` is set. Don't know why.

### DIFF
--- a/plug.vim
+++ b/plug.vim
@@ -99,7 +99,13 @@ let s:mac_gui = has('gui_macvim') && has('gui_running')
 let s:is_win = has('win32')
 let s:nvim = has('nvim-0.2') || (has('nvim') && exists('*jobwait') && !s:is_win)
 let s:vim8 = has('patch-8.0.0039') && exists('*job_start')
-let s:me = resolve(expand('<sfile>:p'))
+if s:is_win && &shellslash
+  set noshellslash
+  let s:me = resolve(expand('<sfile>:p'))
+  set shellslash
+else
+  let s:me = resolve(expand('<sfile>:p'))
+endif
 let s:base_spec = { 'branch': 'master', 'frozen': 0 }
 let s:TYPE = {
 \   'string':  type(''),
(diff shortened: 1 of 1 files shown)


janlazo:
I give up. On Neovim, `set shellslash` with any shell breaks vim-plug on Windows. Vim allows both `shellslash` and `noshellslash` for any shell after this change. Tried using `system([])` to reduce the batchfiles on Neovim but there's no effect. It might be failing on `s:spawn` or earlier.

plug.vim
```diff
@@ -2001,6 +2044,10 @@ function! s:shellesc_ps1(arg)
   return "'".substitute(escape(a:arg, '\"'), "'", "''", 'g')."'"
 endfunction
 
+function! s:shellesc_sh(arg)
+  return "'".substitute(a:arg, "'", "'\\\\''", 'g')."'"
```
janlazo: Pipeline passes and I didn't encounter any issues when running the tests locally or using my config so I assume this is good enough.

plug.vim
```diff
@@ -110,18 +116,50 @@ let s:TYPE = {
 let s:loaded = get(s:, 'loaded', {})
 let s:triggers = get(s:, 'triggers', {})
 
+if s:is_win
+  function! s:plug_call(fn, ...)
+    let shellslash = &shellslash
+    try
+      set noshellslash
```
janlazo: I wondier if this is enough because https://github.com/vim/vim/commit/50f91d22bd81819d5d946f6c38252c7922120ec6 mentions that `completeslash` affects `globpath`. Could it be an issue for fzf.

janlazo:
Neovim works with `set shellslash` if, on `s:spawn`, `set noshellslash` is set temporarily within try/finally/endtry. Why?

ZSaberLv0:
Any further updates? I think `shell=cmd.exe` and `shellslash` is more convenient for cross platform developing, especially for guys like me who are coding both on Windows and *nix using vim as portable solution

janlazo:
@ZSaberLv0 If you upgrade your Vim, you can set `completeslash` instead. If that's insufficient and there is a regression, create a new ticket. `set shell=cmd.exe shellslash` is a landmine because of `shellescape`, `tempname`, and other functions and options that deal with filepaths. Vim and Neovim haven't fixed it for Windows to this day and many 3rd-party plugins break if you use non-default shell settings.

ZSaberLv0:
thanks for your tip, I'll try `completeslash`

Issue #920, opened 2019-12-26
vim-plug requires +iconv but doesn't work with +iconv/dyn


DervishD:
Latest `vim-plug` version require `+iconv` to run shell commands on Windows, but doesn't work when `+iconv/dyn` is available. That didn't happen in the previous version.

This was introduced in #913 and I think it's a good idea, except that for some reason at least in my system, `has(iconv)` returns 0 even though I have `+iconv/dyn` and the `iconv()` function works perfectly within Vim.

Maybe it's something I did wrong in my system, I really don't know, but if I can't fix this I can't use vim-plug and I depend on it somewhat.

Thanks in advance and feel free to ask me for additional information.

------------------------------

```
VIM - Vi IMproved 8.1 (2018 May 18, compiled Nov  1 2019 01:01:21)
MS-Windows 64-bit GUI version
Included patches: 1-2237
Compiled by stop_@LAPTOP-E7OHISR4
Huge version with GUI.  Features included (+) or not (-):
+acl                +clipboard          +directx            +gettext/dyn        +localmap           +packages           +scrollbind         -termguicolors      +visual
+arabic             +cmdline_compl      -dnd                -hangul_input       +lua/dyn            +path_extra         +signs              +terminal           +visualextra
+autocmd            +cmdline_hist       -ebcdic             +iconv/dyn          +menu               +perl/dyn           +smartindent        -termresponse       +viminfo
+autochdir          +cmdline_info       +emacs_tags         +insert_expand      +mksession          +persistent_undo    +sound              +textobjects        +vreplace
+autoservername     +comments           +eval               +job                +modify_fname       -postscript         +spell              +textprop           -vtp
+balloon_eval       +conceal            +ex_extra           +jumplist           +mouse              +printer            +startuptime        -tgetent            +wildignore
-balloon_eval_term  +cryptv             +extra_search       +keymap             +mouseshape         +profile            +statusline         +timers             +wildmenu
+browse             +cscope             -farsi              +lambda             +multi_byte_ime/dyn +python/dyn         -sun_workshop       +title              +windows
++builtin_terms     +cursorbind         +file_in_path       +langmap            +multi_lang         +python3/dyn        +syntax             +toolbar            +writebackup
+byte_offset        +cursorshape        +find_in_path       +libcall            +mzscheme/dyn       +quickfix           +tag_binary         +user_commands      -xfontset
[...]

PR #921, opened 2019-12-27
Use iconv() only if +iconv is enabled.

janlazo:
+iconv or +iconv/dyn are required for iconv().

Close https://github.com/junegunn/vim-plug/issues/920

cc @DervishD

### DIFF
--- a/plug.vim
+++ b/plug.vim
@@ -184,8 +184,8 @@ function! s:define_commands()
   endif
   if !has('nvim')
     \ && (has('win32') || has('win32unix'))
-    \ && (!has('multi_byte') || !has('iconv'))
-    return s:err('Vim needs +iconv, +multi_byte features on Windows to run shell commands.')
+    \ && !has('multi_byte')
+    return s:err('Vim needs +multi_byte feature on Windows to run shell commands. Enable +iconv for best results.')
   endif
   command! -nargs=* -bar -bang -complete=customlist,s:names PlugInstall call s:install(<bang>0, [<f-args>])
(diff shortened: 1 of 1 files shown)


DervishD:
Thanks a lot for the fix, Jan ;)

Still, there's a problem on MY side, not vim-plug's fault, because for some reason my Vim doesn't load libiconv. That means my `iconv()` is only partially working.

Also, that means my `has('iconv')` will return 0 anyway, and as far as I know **there's NO `has('iconv/dyn')`** so your fix has no effect anyway.

This said, the good news (for you) is that really no fix is needed, it was my fault from the beginning, but I discovered it just minutes ago, so I'm very sorry for having reported a non-existing bug.

And again, this said, I think that the code would be cleaner if the tests for `multi_byte` and `iconv` were separated with two different error messages, because right now the end user doesn't know if the problem comes from `iconv` or `multi_byte`, making more difficult to fix the issue. It took me some time to determine my problem was `iconv` and not `multi_byte`.

Other than that, again I'm VERY sorry, Jan.

janlazo:
`+multi_byte` is enabled on build time so it's likely enabled on Windows. It's a sanity check. `+iconv` is the bigger issue because there's no way to easy way to check if `iconv()` works as intended. `iconv()` can work without `+iconv` or `iconv.dll`.

Thanks for bringing this up.

janlazo:
@junegunn Another update for fzf.

DervishD:
Thanks for the fix! Well, it's not a fix, it's really a workaround for systems like mine, with a non-functioning libiconv, but thanks for taking the burden of making vim-plug work in those systems.

janlazo:
It works on my machine :tm: so it's good to go.

PR #930, opened 2020-01-23
Fix fatal invalid reference for repos have no tags

douglarek:
If a repo (such as [is.vim](https://github.com/haya14busa/is.vim)) has no tags , when invoke `:PlugUpdate`, it will occur `fatal invalid reference`.

The patch will skip tags check for this case.

### DIFF
--- a/plug.vim
+++ b/plug.vim
@@ -1158,7 +1158,7 @@ function! s:update_finish()
       if has_key(spec, 'commit')
         call s:log4(name, 'Checking out '.spec.commit)
         let out = s:checkout(spec)
-      elseif has_key(spec, 'tag')
+      elseif has_key(spec, 'tag') && spec.tag
         let tag = spec.tag
         if tag =~ '\*'
           let tags = s:lines(s:system('git tag --list '.plug#shellescape(tag).' --sort -version:refname 2>&1', spec.dir))


janlazo:
Your fix works but the tests fail. Can you update the tests to reflect the new behavior and perhaps use that repo for a new test case?

douglarek:
@janlazo  I am not very familiar with vim script, would you like to help this ?

junegunn:
I don't understand how this helps. Please post the output of `:echo g:plugs`.

douglarek:
`:echo g:plugs` output https://gist.github.com/douglarek/f9355d0e6b2f784c99570dc3f6c8c40a

junegunn:
I see `'tag': ''` all over which is clearly misconfiguration. You instruct vim-plug to use an empty tag, and vim-plug naturally fails with a proper error message because there is no empty tag. So why do you do it?

douglarek:
Just do not know why empty tag added to `g:plugs`, I am using [space-vim](https://github.com/liuchengxu/space-vim).

junegunn:
Well, then it's a bug of space-vim misusing vim-plug. Please report the problem to space-vim. Thanks.

PR #935, opened 2020-02-11
Conditional loading

brglng:
Add an `if` option to load a plugin only if some condition is met. The plugin won't be cleaned by `PlugClean` if the condition is not met.

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,6 +163,7 @@ Reload .vimrc and `:PlugInstall` to install plugins.
 | `dir`                   | Custom directory for the plugin                  |
 | `as`                    | Use different name for the plugin                |
 | `do`                    | Post-update hook (string or funcref)             |
+| `if`                    | Conditional loading                              |
 | `on`                    | On-demand loading: Commands or `<Plug>`-mappings |
 | `for`                   | On-demand loading: File types                    |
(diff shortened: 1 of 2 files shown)


junegunn:
Thanks for your interest in the project, but we try hard to keep the feature set as small as possible so we do not add a new option unless it's absolutely necessary. The wiki page below shows how you can achieve the behavior with the existing options and native VimScript constructs:

https://github.com/junegunn/vim-plug/wiki/tips#conditional-activation

wookayin:
I would much prefer some syntax like
```
Plug 'benekastah/neomake', Cond(has('nvim'))
```
to
```
if has('nvim')
  Plug 'benekastah/neomake', Cond(has('nvim'))
endif
```
because it reads much better and clean, and looks aligned.

https://github.com/junegunn/vim-plug/wiki/tips#conditional-activation would be a nice trick, but the semantics is a little different because such plugins will be installed/registered but just NOT loaded. In fact, such conditionally-not-loaded plugins would be still added to the list `g:plugs`, in which case one cannot simply tell whether the plugin is available or not for further customization/configuration. 

So I think adding the conditional loading feature (`if` syntax) would provide much flexibility.

PR #936, opened 2020-02-18
Guard against empty tags

janlazo:
Related: https://github.com/junegunn/vim-plug/pull/930

Perhaps throwing early is safer than proceeding with invalid values.

### DIFF
--- a/plug.vim
+++ b/plug.vim
@@ -623,7 +623,7 @@ function! plug#(repo, ...)
 
   try
     let repo = s:trim(a:repo)
-    let opts = a:0 == 1 ? s:parse_options(a:1) : s:base_spec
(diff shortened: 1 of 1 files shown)


junegunn:
Hmm, I don't know, are we going to sanity check every other option? i.e. `dir`, `branch`, `commit`, ...

#930 was caused by misbehavior of a configuration framework (which is [now fixed](https://github.com/liuchengxu/space-vim/commit/dead5e868bf1daa4d7a49d190be665d86ffbcdfd)), and it's very unlikely someone will actually make such a mistake.

junegunn:
> Perhaps throwing early is safer than proceeding with invalid values.

If we're going to do this, this is how I think we should do it, giving the user a chance to correct their configuration.

janlazo:
I'll create a new PR to throw instead of continuing with invalid values.

PR #955, opened 2020-04-04
improved path creation in powershell readme example

fourtf:
I have removed the `mkdir` alias in powershell which breaks the example in the readme since `md` seems to execute `mkdir`. Furthermore using full function names, e.g. `New-Item` instead of md, is preferred so situations such as this do not happen. I have also replaced the `GetUnresolvedProviderPathFromPSPath` call with using the `USERPROFILE` environment variable.

Note that `-Force` for directories makes sure that _all_ missing directories are created and does __not__ delete existing directories.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,13 +44,11 @@ file as suggested [here][auto].
 ###### Windows (PowerShell)
 
 ```powershell
-md ~\vimfiles\autoload
+New-Item ${env:USERPROFILE}\vimfiles\autoload -ItemType Directory -Force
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -44,13 +44,11 @@ file as suggested [here][auto].
 ###### Windows (PowerShell)
 
 ```powershell
-md ~\vimfiles\autoload
+New-Item ${env:USERPROFILE}\vimfiles\autoload -ItemType Directory -Force
```
janlazo, line 47: You didn't escape this.

janlazo:
Resolved in https://github.com/junegunn/vim-plug/pull/1003.

Issue #852, opened 2019-06-25
Error when I try to install plugins on Windows


jdhao:
I am using vim-plug to manage my plugins both on Windows and Linux. But on Windows, I encounter errors sometimes when I try to install new plugins using `PlugInstall` command. The error is shown blow:

```
|| Error detected while processing function
C:\Users\Administrator\AppData\Local\nvim\autoload\plug.vim|604| <SNR>2_install[1]
C:\Users\Administrator\AppData\Local\nvim\autoload\plug.vim|1015| <SNR>2_update_impl[58]
C:\Users\Administrator\AppData\Local\nvim\autoload\plug.vim|327| <SNR>2_git_version_requirement[2]
C:\Users\Administrator\AppData\Local\nvim\autoload\plug.vim|2039| <SNR>2_system[6]
|| E482: Can't open file C:\Users\ADMINI~1\AppData\Local\Temp\nvimBCqYQf\10.bat for writing: no such file or directory
|| Error detected while processing function
C:\Users\Administrator\AppData\Local\nvim\autoload\plug.vim|604| <SNR>2_install[1]
C:\Users\Administrator\AppData\Local\nvim\autoload\plug.vim|1015| <SNR>2_update_impl[58]
C:\Users\Administrator\AppData\Local\nvim\autoload\plug.vim|327| <SNR>2_git_version_requirement[2]
|| E171: Missing :endif
```

This error seems random and only appear after I have used nvim-qt for some time. If I quit nvim-qt and restart, I can use `PlugInstall` without any error.

## version info

```
NVIM v0.4.0-819-g66d127957
Build type: RelWithDebInfo
LuaJIT 2.0.5
[...]

PR #956, opened 2020-04-04
Support list type command for s:system to reduce batchfiles on Windows

janlazo:
`s:system` should be safe to optimize because some commands are internal and don't rely on the shell.

Any perf changes on non-Windows OS should be negligible. Neovim supports list type command for `system` so Neovim users can expect improved performance, at least on Windows.

Related: https://github.com/junegunn/vim-plug/issues/690, https://github.com/junegunn/vim-plug/issues/852

### DIFF
--- a/plug.vim
+++ b/plug.vim
@@ -2069,6 +2069,9 @@ endfunction
 "          If unset, fallback to 'cmd.exe' on Windows or 'sh'.
 " - script: If truthy and shell is cmd.exe, escape for batchfile syntax.
 function! plug#shellescape(arg, ...)
+  if a:arg =~# '^[A-Za-z0-9_/:.-]\+$'
+    return a:arg
+  endif
   let opts = a:0 > 0 && type(a:1) == s:TYPE.dict ? a:1 : {}
   let shell = get(opts, 'shell', s:is_win ? 'cmd.exe' : 'sh')
   let script = get(opts, 'script', 1)
--- a/plug.vim
+++ b/plug.vim
@@ -1302,10 +1302,12 @@ function! s:spawn(name, cmd, opts)
     endif
   elseif s:vim8
     let cmd = has_key(a:opts, 'dir') ? s:with_cd(a:cmd, a:opts.dir, 0) : a:cmd
-    let argv = s:is_win ? ['cmd', '/s', '/c', '"'.cmd.' 2>&1"'] : ['sh', '-c', cmd.' 2>&1']
+    let argv = s:is_win ? ['cmd', '/s', '/c', '"'.cmd.'"'] : ['sh', '-c', cmd]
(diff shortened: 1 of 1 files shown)


plug.vim
```diff
@@ -2185,7 +2218,9 @@ endfunction
 
 function! s:rm_rf(dir)
   if isdirectory(a:dir)
-    call s:system((s:is_win ? 'rmdir /S /Q ' : 'rm -rf ') . plug#shellescape(a:dir))
+    call s:system(s:is_win
+    \ ? 'rmdir /S /Q '.plug#shellescape(a:dir)
```
junegunn, line 2222: Should we change this one to an array as well?

janlazo: `rmdir` is a cmd.exe builtin command. I don't know a default external program for deleting files/directories on Windows. If there is a dll for this, then we can replace this line with `['rundll32.exe', 'delete.dll,DeleteProtocol', a:dir]`.

janlazo:
> Is the performance improvement noticeable?

It shaves a couple of seconds when testing this PR with my config.

plug.vim
```diff
@@ -1411,7 +1425,7 @@ while 1 " Without TCO, Vim stack is bound to explode
     endif
   else
     call s:spawn(name,
-          \ printf('git clone %s %s %s %s 2>&1',
+          \ printf('git clone %s %s %s %s',
           \ has_tag ? '' : s:clone_opt,
```
janlazo, line 1429: `s:clone_opt` is the only blocker for using list type for the command passed to `s:spawn`. Using list type command lets Vim/Neovim run git directly without the shell. I'll refactor `s:clone_opt` to a list in separate PR since it affects the Python and Ruby code.

janlazo:
No issues when testing this patch with my config. Merging later.

plug.vim
```diff
@@ -1273,12 +1280,15 @@ function! s:spawn(name, cmd, opts)
   let job = { 'name': a:name, 'running': 1, 'error': 0, 'lines': [''],
             \ 'new': get(a:opts, 'new', 0) }
   let s:jobs[a:name] = job
-  let cmd = has_key(a:opts, 'dir') ? s:with_cd(a:cmd, a:opts.dir, 0) : a:cmd
-  let argv = s:is_win ? ['cmd', '/s', '/c', '"'.cmd.'"'] : ['sh', '-c', cmd]
 
   if s:nvim
+    if has_key(a:opts, 'dir')
+      let job.cwd = a:opts.dir
```
janlazo, line 1286: This should work on Vim 8 as well but CI will fail on Ubuntu bionic and latest vim version.

janlazo:
Next is `s:bang`. Neovim does not support working directory for `:!` commands, unlike system.

PR #961, opened 2020-04-26
:Plug throws error for invalid option

janlazo:
Related: #930, #936

### DIFF
--- a/plug.vim
+++ b/plug.vim
@@ -632,17 +632,34 @@ function! plug#(repo, ...)
     let g:plugs[name] = spec
     let s:loaded[name] = get(s:loaded, name, 0)
   catch
-    return s:err(v:exception)
+    return s:err(repo . ' ' . v:exception)
   endtry
 endfunction
 
 function! s:parse_options(arg)
   let opts = copy(s:base_spec)
   let type = type(a:arg)
+  let opt_errfmt = 'Invalid argument for "%s" option of :Plug (expected: %s)'
   if type == s:TYPE.string
+    if empty(a:arg)
+      throw printf(opt_errfmt, 'tag', 'string')
+    endif
     let opts.tag = a:arg
   elseif type == s:TYPE.dict
     call extend(opts, a:arg)
+    for opt in ['branch', 'tag', 'commit', 'rtp', 'dir', 'as']
+      if has_key(opts, opt)
+      \ && (type(opts[opt]) != s:TYPE.string || empty(opts[opt]))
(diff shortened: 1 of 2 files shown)


plug.vim
```diff
@@ -632,17 +632,25 @@ function! plug#(repo, ...)
     let g:plugs[name] = spec
     let s:loaded[name] = get(s:loaded, name, 0)
   catch
-    return s:err(v:exception)
+    return s:err(repo . ' ' . v:exception)
```
janlazo, line 635: Hard to debug which `:Plug` caused the error without this.

janlazo:
@junegunn Any issues with the implementation?

junegunn:
Looks good to me.

1. We can also verify `do` option value.
2. Does this affect the startup time? Would it be negligible given that we have ten to twenty plugins with dictionary options?

janlazo:
1. `Verifying 'do' at `Plug` breaks existing tests. Expected error message is different from the error format I am using for the new tests I added.

2. Is 2ms negligible? I tested with my config (about 50 plugins, 9 plugins passing a dictionary) and I don't notice a slowdown.

test/workflow.vader
```diff
@@ -1055,15 +1059,13 @@ Execute (Post-update hook output; success and failure):
...
-Execute (Post-update hook output; invalid type or funcref):
+Execute (Post-update hook output; invalid funcref):
   call plug#begin()
-  Plug 'junegunn/vim-easy-align', { 'do': 1 }
   Plug 'junegunn/vim-pseudocl',   { 'do': function('call') }
   call plug#end()
 
   silent PlugInstall! 1
   AssertEqual 'x Post-update hook for vim-pseudocl ... Vim(call):E119: Not enough arguments for function: call', getline(5)
-  AssertEqual 'x Post-update hook for vim-easy-align ... Invalid hook type', getline(6)
```
janlazo, line 1046: `Invalid hook type` can happen if the user updates `g:plugs` after `plug#end` and before `PlugInstall`.

junegunn:
```sh
#!/bin/bash

niter=100

rm -f plug.log

measure() {
  for _ in $(seq 1 $niter); do
    vim --startuptime _.log +q
    tail -1 _.log >> plug.log
  done
  cat plug.log
  awk '{ sum += $1 } END { printf "mean: %dms\n", sum / NR }' < plug.log
}

measure
```

Thanks, I also see no noticeable difference in startup time before and after the change.

PR #1001, opened 2020-08-02
Fix typos

rmschindler:


README.md
```diff
@@ -7,7 +7,7 @@ A minimalist Vim plugin manager.
 
 ### Pros.
 
-- Easier to setup: Single file. No boilerplate code required.
+- Easier to set up: Single file. No boilerplate code required.
```
janlazo, line 10: What about `Easy to set up:`? `ier` implies that there is a comparison with another package manager but it is not mentioned in the README.

Next line also uses `Easier` which should be `Easy`.

rmschindler: Yes, you are right. I agree that 'easy' is better than 'easier.'

janlazo:
Thanks.

PR #1003, opened 2020-08-06
Update README.md (Powershell Install One-Liner)

uciti:
A tidy one liner for the powershell install command. 
Much like the unix one 😄

Describe the details of your PR ...

This just makes the powershell install command a short one-liner.
- `iwr -useb` works on any version of powershell AFAIK.
- `ni` is `New-Item` and accepts content from `iwr` via the pipe. and the `-Force` on `ni` creates directories if they don't exist.
- the single backtick is the powershell equivalent of the backslash in unix for escaping the newline.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,14 +44,8 @@ file as suggested [here][auto].
 ###### Windows (PowerShell)
 
 ```powershell
-md ~\vimfiles\autoload
-$uri = 'https://raw.githubusercontent.com/junegunn/vim-plug/master/plug.vim'
-(New-Object Net.WebClient).DownloadFile(
-  $uri,
-  $ExecutionContext.SessionState.Path.GetUnresolvedProviderPathFromPSPath(
-    "~\vimfiles\autoload\plug.vim"
-  )
-)
+iwr -useb https://raw.githubusercontent.com/junegunn/vim-plug/master/plug.vim |`
+    ni $HOME/vimfiles/autoload/plug.vim -Force
 ```
 
 #### Neovim
@@ -73,14 +67,8 @@ curl -fLo ~/.var/app/io.neovim.nvim/data/nvim/site/autoload/plug.vim \
 ###### Windows (PowerShell)
 
 ```powershell
-md ~\AppData\Local\nvim-data\site\autoload
-$uri = 'https://raw.githubusercontent.com/junegunn/vim-plug/master/plug.vim'
-(New-Object Net.WebClient).DownloadFile(
-  $uri,
-  $ExecutionContext.SessionState.Path.GetUnresolvedProviderPathFromPSPath(
-    "~\AppData\Local\nvim-data\site\autoload\plug.vim"
-  )
-)
+iwr -useb https://raw.githubusercontent.com/junegunn/vim-plug/master/plug.vim |`
+    ni $HOME/AppData/Local/nvim-data/site/autoload/plug.vim -Force
 ```
 
 ### Getting Help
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -44,14 +44,8 @@ file as suggested [here][auto].
...
 ```powershell
-md ~\vimfiles\autoload
-$uri = 'https://raw.githubusercontent.com/junegunn/vim-plug/master/plug.vim'
-(New-Object Net.WebClient).DownloadFile(
-  $uri,
-  $ExecutionContext.SessionState.Path.GetUnresolvedProviderPathFromPSPath(
-    "~\vimfiles\autoload\plug.vim"
-  )
-)
+iwr -useb https://raw.githubusercontent.com/junegunn/vim-plug/master/plug.vim |`
```
janlazo, line 47: The lines ends with a pipe, `|`. Does the backtick make a difference on old powershell versions?

README.md
```diff
@@ -73,14 +67,8 @@ curl -fLo ~/.var/app/io.neovim.nvim/data/nvim/site/autoload/plug.vim \
...
-md ~\AppData\Local\nvim-data\site\autoload
-$uri = 'https://raw.githubusercontent.com/junegunn/vim-plug/master/plug.vim'
-(New-Object Net.WebClient).DownloadFile(
-  $uri,
-  $ExecutionContext.SessionState.Path.GetUnresolvedProviderPathFromPSPath(
-    "~\AppData\Local\nvim-data\site\autoload\plug.vim"
-  )
-)
+iwr -useb https://raw.githubusercontent.com/junegunn/vim-plug/master/plug.vim |`
+    ni $HOME/AppData/Local/nvim-data/site/autoload/plug.vim -Force
```
janlazo, line 71:
```powershell
ni "$env:LOCALAPPDATA/nvim-data/site/autoload/plug.vim" -Force
```

README.md
uciti: The backtick is the standard escape character for every version of powershell i've used v2.0/v5.0/v7.0

This was just to make it visually look like the other commands for unix. It could also just be a single one-liner with no break.

README.md
uciti: good catch!

README.md
janlazo: That's not what I meant. If a command ends with a pipe, then the shell won't execute the command immediately.

```powershell
> echo a |
>> cat
a
```

`cat` is given on the nested prompt at the next line. `echo a |` is incomplete so `echo a` does not run.

Good to know that backtick works in Powershell v2.

uciti: Ah ok, I misunderstood, my bad.

Yes the backtick does escape the command.

It should be good either way. Especially for any version of PowerShell running by default on windows 7/8/10.

README.md
```diff
@@ -44,14 +44,8 @@ file as suggested [here][auto].
...
-md ~\vimfiles\autoload
-$uri = 'https://raw.githubusercontent.com/junegunn/vim-plug/master/plug.vim'
-(New-Object Net.WebClient).DownloadFile(
-  $uri,
-  $ExecutionContext.SessionState.Path.GetUnresolvedProviderPathFromPSPath(
-    "~\vimfiles\autoload\plug.vim"
-  )
-)
+iwr -useb https://raw.githubusercontent.com/junegunn/vim-plug/master/plug.vim |`
+    ni $HOME/vimfiles/autoload/plug.vim -Force
```
janlazo, line 48: There should be double-quotes here because of whitespace.

PR #1006, opened 2020-08-26
Add example of how to install plugin using SHA

ayushxx7:
- No tests have been ran as this is a change in the README.md file.

Describe the details of your PR 
- Just added an example of how to install a plugin via vim-plug using commit SHA.


junegunn:
Thanks, but I think it's enough that we mention `commit` option along with `branch` and `tag` in https://github.com/junegunn/vim-plug#plug-options as `commit` option is rarely used in practice.

The reason I don't want to add more examples is two-fold. 1. I'd like to keep the documentation short and concise as much as possible. 2. I've noticed that some users blindly copy-and-paste these examples to their configs. It's their fault, but I'm just concerned about the possibility.

PR #1007, opened 2020-08-27
Support non-master default branch

mattn:
https://github.blog/changelog/2020-08-26-set-the-default-branch-for-newly-created-repositories/

GitHub will be change the default branch to `main` for newly created repository. vim-plug use `master` branch for the default. This change look current branch name from HEAD/ref.

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,7 +116,7 @@ Plug 'SirVer/ultisnips' | Plug 'honza/vim-snippets'
 Plug 'scrooloose/nerdtree', { 'on':  'NERDTreeToggle' }
 Plug 'tpope/vim-fireplace', { 'for': 'clojure' }
 
-" Using a non-master branch
+" Using a non-default branch
 Plug 'rdnetto/YCM-Generator', { 'branch': 'stable' }
 
 " Using a tagged release; wildcard allowed (requires git 1.9.2 or above)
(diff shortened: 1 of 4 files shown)


mattn:
This change add number of calls of git command. I suggest to merge #937

https://github.com/junegunn/vim-plug/pull/937

junegunn:
Thanks, but this PR is missing `s:git_get_branch`.

mattn:
@junegunn Thank you. What still wrong?

https://travis-ci.org/github/junegunn/vim-plug/jobs/721704596

junegunn:
Failing test cases are marked with `(X)`.

We have a very comprehensive set of test cases testing every aspect of vim-plug. And obviously this pull request breaks existing test cases written with some assumptions of how vim-plug should behave. We would have to also update existing test cases accordingly.

junegunn:
Let me see if I can fix them.

junegunn:
I've managed to make the existing test cases pass, though I'm not 100% sure if reading `.git/refs/remotes/origin/HEAD` is the correct way to get the default branch of origin. Please review the updated code.

mattn:
Thank you. Looks good to me.

leitzler:
Thanks for this PR @mattn! As far as I understand, vim-plug will now fetch the default branch when adding a new plugin.

Since a lot of repositories also follow GitHubs rename instructions I suggest that vim-plug also run `git remote set-head origin -a` each update to ensure that the remote default branch isn't stale.

Thoughts?

mattn:
I'm okay. Any thought? @junegunn

PR #1026, opened 2020-10-19
Disable credential-helper

mattn:
When I start `:PlugUpdate` on WIndows, some fatal errors occured. And it seems failing lock the credentail store.

https://stackoverflow.com/questions/37505843/fatal-unable-to-get-credential-storage-lock-file-exists/37907588

Certainly, I use `credential.help=store` in global config. but it is needed for others. So I add `-c credential.helper=""` to disable it on vim-plug.

### DIFF
--- a/plug.vim
+++ b/plug.vim
@@ -1078,7 +1078,7 @@ function! s:checkout(spec)
   let output = s:git_revision(a:spec.dir)
   if !empty(output) && !s:hash_match(sha, s:lines(output)[0])
     let output = s:system(
-          \ 'git fetch --depth 999999 && git checkout '.plug#shellescape(sha).' --', a:spec.dir)
+          \ 'git -c credential.helper="" fetch --depth 999999 && git checkout '.plug#shellescape(sha).' --', a:spec.dir)
   endif
   return output
 endfunction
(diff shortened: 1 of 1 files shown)


mattn:
Good point: PlugUpdate seems to be little faster.

plug.vim
```diff
@@ -1078,7 +1078,7 @@ function! s:checkout(spec)
   let output = s:git_revision(a:spec.dir)
   if !empty(output) && !s:hash_match(sha, s:lines(output)[0])
     let output = s:system(
-          \ 'git fetch --depth 999999 && git checkout '.plug#shellescape(sha).' --', a:spec.dir)
+          \ 'git -c credential.helper="" fetch --depth 999999 && git checkout '.plug#shellescape(sha).' --', a:spec.dir)
```
janlazo, line 1081: Are the double quotes necessary?

mattn: If @junegunn  don't want, I will remove them.

junegunn: Let's remove them and worry less about escaping things.

junegunn:
Merged, thanks.

I wonder if there's anyone actually relying on a credential helper setting for installing Vim plugins from private repositories, but I guess it's highly unlikely. Let's do it and see if anyone complains. 🤷

cbows:
I think I might have a complaint, but not for the reason of private repositories. See https://github.com/junegunn/vim-plug/issues/1047

UnlikelyCrusader:
I use plugins across a proxy, so I had to revert these changes.  I would nice if this feature was a configurable setting.

PR #1071, opened 2021-02-08
Better resolving of relative plugin dirs

Roy-Orbison:
When using a relative path such as `'dir': 'some-repo-before-compat-break'` in a `Plug` spec, the folder gets resolved relative to Vim's current directory, which could be anywhere, often the root of a project. I believe it would be far better if this was resolved as relative to `g:plug_home`, so absolute and home-relative `dir` specs remain unchanged, but simple directory names would be in the same place other plugins are kept. This would also mean such plugins are loaded when Vim is opened in another working directory, and can be cleaned away if removed from `.vimrc`.

This works for single level names where the trailing slash is omitted.

### DIFF
--- a/plug.vim
+++ b/plug.vim
@@ -750,7 +750,9 @@ function! s:parse_options(arg)
     endif
     call extend(opts, a:arg)
     if has_key(opts, 'dir')
-      let opts.dir = s:dirpath(s:plug_expand(opts.dir))
+      exe 'cd' fnameescape(g:plug_home)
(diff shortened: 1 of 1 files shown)


junegunn:
It looks like `as` option is what you're looking for. Have you considered using it instead of `dir`?

I agree that resolving `dir` relative to the current working directory is strange but `dir` option was added to support exceptional use cases where the user wants to install the plugin outside of `g:plug_home`.

https://github.com/junegunn/vim-plug/commit/f58d090bb263fdf79d0131f18df29598f7e1b77f

Roy-Orbison:
Right you are.

PR #1080, opened 2021-03-20
Failed to connect to github.com port 443: Timeout

jiangjianshan:
Due to github.com is not stable for some contries, e.g. China. So change it to mirror site instead of direct site. This will solve the issue of “Failed to connect to github.com port 443: Timeout".

### DIFF
--- a/plug.vim
+++ b/plug.vim
@@ -769,7 +769,7 @@ function! s:infer_properties(name, repo)
       if repo !~ '/'
         throw printf('Invalid argument: %s (implicit `vim-scripts'' expansion is deprecated)', repo)
       endif
-      let fmt = get(g:, 'plug_url_format', 'https://git::@github.com/%s.git')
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks for the pull request. I'm glad to hear that you found a solution that works for you. But I'm not sure if it's the right call to use a mirror site as the default at the moment. Actually, this is the first time I've heard about fastgit.org and I'm not confident that it will be properly maintained in the long run.

I suggest that you specify the full URL of the mirror in your configuration file. e.g.

```vim
Plug 'https://hub.fastgit.org/junegunn/gv.vim.git'
```

Or you could define a separate command for converting the URL.

```vim
command! -nargs=+ PlugMirror call call('plug#', [printf('https://hub.fastgit.org/%s.git', [<args>][0])] + [<args>][1:])

PlugMirror 'junegunn/gv.vim'
```

jiangjianshan:
@junegunn , thanks for you suggestion.

PR #1092, opened 2021-05-05
Update README w/ working curl cmd

csrunner20:
Updated the curl cmd to include the -k flag to ignore SSL certification. Idk if that was your intended but w/o the flag you'll get:

```
curl: (60) SSL certificate problem: unable to get local issuer certificate
More details here: https://curl.haxx.se/docs/sslcerts.html

curl failed to verify the legitimacy of the server and therefore could not
establish a secure connection to it. To learn more about this situation and
how to fix it, please visit the web page mentioned above.
```

Not sure if that's a github thing or what, but just for future reference.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ and put it in the "autoload" directory.
 ###### Unix
 
 ```sh
-curl -fLo ~/.vim/autoload/plug.vim --create-dirs \
+curl -kfLo ~/.vim/autoload/plug.vim --create-dirs \
     https://raw.githubusercontent.com/junegunn/vim-plug/master/plug.vim
 ```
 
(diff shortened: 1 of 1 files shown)


junegunn:
Thanks for your interest in the project.

The command should work just fine. There are cases where you have to put `-k` (`--insecure`) to the command (e.g. behind a corporate proxy), but we shouldn't recommend users to use the option in normal circumstances.

See https://github.com/junegunn/vim-plug/wiki/tips#http-proxy-issues-warning

csrunner20:
Ah! Sorry didn't see that tips page! And yes I was on corporate network installing; didn't realize that was what was causing that. 

Sorry for taking up your time! Have a good day Junegunn :D

PR #1110, opened 2021-07-18
Allow command modifiers for on-demand-loading commands

subnut:
Before this commit, commands like this -
```vim
:tab Git diff HEAD^ HEAD
```
were interpreted like this -
```vim
:Git diff HEAD^ HEAD
```
This commit fixes that issue.

### DIFF
--- a/plug.vim
+++ b/plug.vim
@@ -399,7 +399,7 @@ function! plug#end()
 
   for [cmd, names] in items(lod.cmd)
     execute printf(
-    \ 'command! -nargs=* -range -bang -complete=file %s call s:lod_cmd(%s, "<bang>", <line1>, <line2>, <q-args>, %s)',
+    \ 'command! -nargs=* -range -bang -complete=file %s call s:lod_cmd(%s, "<bang>", <line1>, <line2>, <q-args>, <q-mods> ,%s)',
     \ cmd, string(cmd), string(names))
   endfor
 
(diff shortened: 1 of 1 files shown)


tomtomjhj:
I think this needs a version guard because `<mods>` is a relatively new feature ([v7.4.1898](https://github.com/vim/vim/commit/63a60ded3fd584847a05dccf058026e682abad90))

subnut:
@tomtomjhj 2016 is _relatively new_? For whom?  
I'm not trying to be sarcastic... I am genuinely interested in knowing use-cases where people need to use vim versions as old as 2016.

---

TBH, I'd completely forgotten about this PR 😅

tomtomjhj:
Of course, I believe 2016 feature is old enough to use without version check in 2021.
The only reason I think 2016 feature is relatively new is that vim-plug has checks for even older stuff, e.g.
* vim 7.3.442, 2012 https://github.com/junegunn/vim-plug/blob/c9971346bb486ccff354aaee8606cc0d2d5a0c97/plug.vim#L589
* git 2.0 2014 https://github.com/junegunn/vim-plug/commit/2f4e28161e114cc4b34a9b8ff0f51f4906f37b0b

janlazo:
I'm one of those people with old Vim versions so add a version,feature checks until @junegunn changes his mind on the minimum Vim/Neovim versions.

subnut:
@janlazo Will do (please wait a few days, I am currently AFK).


I am curious about why you are still using such an old version of vim.  
Can you please explain? 😅

junegunn:
One of the strong selling points of vim-plug is its excellent backward compatibility. Most of its features work as expected even with Vim 7.0 which means that it works virtually everywhere. There can be various reasons you can't or won't use the latest versions of Vim (company guidelines, security requirements, systems with tiny resources, etc.), and still, you don't have to worry about vim-plug not working. That's a great thing and we're not going to lose that.

janlazo:
@subnut Old company servers. Difficult to install or upgrade dev packages, even if it's just a newer version of Vim. Some of my colleagues don't know or care that vi,vim,`$EDITOR` is any version of vim. I doubt anybody there is aware of `:terminal` command. Some OS packages of vim disable some features are missing a few patches that I need to reliably use the new features (ie. `:packadd`).

subnut:
Tried implementing the version guard in 82ffd4f  
Please check if the implementation is correct

subnut:
@janlazo @junegunn What is blocking this PR from being merged?

subnut:
If there's anything that needs to be done here, please let me know.

junegunn:
This slipped my mind. Thanks for your patience. I'll review the patch.

junegunn (approved):

junegunn:
Works great, thanks for the patch! Sorry for the long delay.

PR #1117, opened 2021-08-02
Set --origin=origin for git clone commands

gibfahn:
---

#### Commits _(oldest to newest)_

c78dc4d Set --origin=origin for git clone commands

Otherwise if the user has set a `git config clone.defaultRemoteName
foo`, then vim-plug will fail to detect the latest upstream changes as
the remote will be incorrect, and will repeatedly state that the plugin
repo needs to be cleaned.

<br/>

### DIFF
--- a/plug.vim
+++ b/plug.vim
@@ -1208,7 +1208,8 @@ function! s:update_impl(pull, force, args) abort
   normal! 2G
   silent! redraw
 
-  let s:clone_opt = []
+  " Set remote name, overriding a possible user git config's clone.defaultRemoteName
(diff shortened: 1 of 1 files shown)


bobrippling:
I've some partially complete local work that does something similar, but lets you set the remote for `:PlugUpdate`, etc.

This is because locally I'll have a few remotes - the upstream and my own fork, and possibly others, and when migrating to `vim-plug`, I didn't want to rename specific ones to `origin` for it to work.

Anyone got thoughts on that use-case?

junegunn (approved):
LGTM. Thanks.

PR #1151, opened 2022-01-02
README: clarify plugins directory for Windows Vim users

shawnhatori:
The user runtime directory for Vim on Windows is `~/vimfiles` by default, not `~/.vim`. While this doesn't actually affect the functionality of vim-plug, this change would indicate to Windows users that the vim-plug directory for plugins should correspond to this, rather than erroneously creating a new `.vim` folder with only `/plugged` in it. I had made this mistake myself before fixing it manually.

This change would be consistent with the installation instructions for Windows, which installs vim-plug into `~/vimfiles`:
https://github.com/junegunn/vim-plug#windows-powershell

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,6 +97,7 @@ Add a vim-plug section to your `~/.vimrc` (or `stdpath('config') . '/init.vim'` 
 
 ```vim
 " Specify a directory for plugins
+" - For Windows: '~/vimfiles/plugged'
 " - For Neovim: stdpath('data') . '/plugged'
 " - Avoid using standard Vim directory names like 'plugin'
 call plug#begin('~/.vim/plugged')
--- a/README.md
+++ b/README.md
@@ -97,8 +97,9 @@ Add a vim-plug section to your `~/.vimrc` (or `stdpath('config') . '/init.vim'` 
 
 ```vim
 " Specify a directory for plugins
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -97,6 +97,7 @@ Add a vim-plug section to your `~/.vimrc` (or `stdpath('config') . '/init.vim'`
 
 ```vim
 " Specify a directory for plugins
+" - For Windows: '~/vimfiles/plugged'
```
junegunn, line 100: This doesn't apply to Neovim on Windows I guess? If that is the case, we should put this line *after* `For Neovim`.

shawnhatori: That's right, it doesn't apply to Neovim on Windows.

Neovim handles this problem in a much more sensible way, since it has `stdpath('data')` to generate the correct path for all platforms. The existing directions for Neovim are correct for Linux/macOS/Windows.

So with this change, the order listed in the README would be:
Vim (Linux/macOS)
Vim (Windows)
Neovim (all platforms)

I'm happy to rearrange the order or make the text more explicit if that would be preferred.

shawnhatori: I've pushed a new commit consistent with my comment above.

junegunn (approved):

junegunn:
Merged, thanks!

Issue #1175, opened 2022-04-20
The check for did_load_filetypes is too strict


lervag:
Neovim v0.7 introduces the opt-in `filetype.lua` to replace `filetype.vim` possibly in v0.8. To use this and disable `filetype.vim`, one is suggested to add

```vim
let g:do_filetype_lua = 1
let g:did_load_filetypes = 0
```

to ones `init.vim` file. If you add this before `plug#end()` is called, then vim-plug will unset `g:did_load_filetypes` due to this code:

https://github.com/junegunn/vim-plug/blob/93ab5909784e09134e90f15cafa8a5edcc9a00fe/plug.vim#L355-L357

I suggest changing the conditional to this:

```vim
if exists('g:did_load_filetypes') && g:did_load_filetypes > 0
  filetype off
endif
```

See e.g. [@gpanders's blog post](https://gpanders.com/blog/whats-new-in-neovim-0-7/#filetypelua) for some more writeup on this particular feature in neovim.

Shougo:
I think this is better.

`if get('g:', 'did_load_filetypes', 0)`

junegunn:
Thanks. Would you be willing to submit a pull request?

lervag:
Yes, sure. I'm closing this as it is addressed in #1176.

PR #1176, opened 2022-04-21
fix: dont "filetype off" unless did_load_filetypes

lervag:
This resolves the issue described in #1175.

### DIFF
--- a/plug.vim
+++ b/plug.vim
@@ -352,7 +352,7 @@ function! plug#end()
   endif
   let lod = { 'ft': {}, 'map': {}, 'cmd': {} }
 
-  if get(g:, 'did_load_filetypes', 0) > 0
+  if get(g:, 'did_load_filetypes', 0)
     filetype off
   endif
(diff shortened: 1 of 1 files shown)


plug.vim
```diff
@@ -352,7 +352,7 @@ function! plug#end()
   endif
   let lod = { 'ft': {}, 'map': {}, 'cmd': {} }
 
-  if exists('g:did_load_filetypes')
+  if get(g:, 'did_load_filetypes', 0) > 0
```
junegunn, line 355: Let's drop `> 0` as @Shougo initially suggested.

junegunn (changes requested):

plug.vim
lervag: Done!

junegunn:
Merged, thanks!

lervag:
Great! Thanks for providing this excellent plugin manager! I've been a happy user since way back :D

PR #1204, opened 2022-08-17
Add gitignore so that vim-plug can be managed by vim-plug

smuenzel:
If you set up vim-plug to manage vim-plug itself, it will create a doc/tags file, which means the repo it is saved in is dirty. Fix this by adding a gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+doc/tags


junegunn:
I have to mention this:

* https://github.com/junegunn/vim-plug/wiki/faq#shouldnt-vim-plug-update-itself-on-plugupdate-like-vundle
* https://github.com/junegunn/vim-plug/pull/240#issuecomment-110538613

But, having a .gitignore file wouldn't hurt.

PR #1215, opened 2022-10-08
Fixed Neovim installation documentation.

DBragz:
Was trying to install vim-plug into Neovim and the initial installation script did not work. After reviewing a bit about Neovim I found the proper default location to place the plug.vim file. No code is being checked in just documentation. Thanks!


DBragz:
Does not need to be updated.

PR #1223, opened 2022-12-19
Add a number of public utility functions

rsrchboy:
Some trivial functions to access `s:loaded`, to allow sane introspection of plugin status, etc.

### DIFF
--- a/plug.vim
+++ b/plug.vim
@@ -2808,5 +2808,17 @@ if exists('g:plugs')
   call s:define_commands()
 endif
 
+function! plug#plugin_info(name) abort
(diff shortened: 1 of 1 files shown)


Arnie97:
Wouldn't `!plug#loaded()` be sufficient to know a plugin is `not_loaded()`?
I don't see the need for two functions.

Kamilcuk:
I vote for this! What I would really want `plug#is_loaded(name)` so that I can check if a plugin is loaded and have plugin specific config there.

junegunn:
What's the exact use case? Maybe some pseudo-code of what you're trying to achieve with these?

PR #1255, opened 2023-09-26
Add reload vimrc step to usage section

DanielApt:
Reloading the .vimrc should be part of the usage section, not the example section. 

As a new user, I overlooked that reloading `~/.vimrc` was needed. [Take a look at the _Usage_ section](https://github.com/junegunn/vim-plug#usage) and see how it's easy to overlook the source and `:PlugInstall` steps.

I believe this update makes it much clearer.


junegunn:
Merged, thanks! I changed `.vimrc` to "the file" because it's `init.vim` in Neovim.

Issue #1005, opened 2020-08-13
Use the default branch instead of master when branch isn't set explicitly


leitzler:
[As several repositories are already moving away](https://github.com/github/renaming) from having the default branch named `master`, I suggest that `vim-go` should use the repository default branch when the user hasn't explicitly set a branch.

I'm using https://github.com/govim/govim (installed via `vim-plug`) that recently changed from `master` to `main` as the default branch. A workaround to get things working was to rename the local branch inside `.vim/plugged/govim` but I recently noticed that `vim-plug` changes branch back to `master` if I don't have a branch name set.

i.e.
```
" This will give me the default branch instead of 'master'
Plug 'govim/govim' 

" .. while this still works
Plug 'govim/govim', { 'branch': 'development' }
```

------------------------------

```
VIM - Vi IMproved 8.2 (2019 Dec 12, compiled Jun  7 2020 01:33:04)
macOS version
Included patches: 1-916
Compiled by leitzler@machine.int.leitzler.net
Huge version without GUI.  Features included (+) or not (-):
+acl               +cryptv            +fork()            +modify_fname      +persistent_undo   +syntax            +visualextra
+arabic            -cscope            -gettext           +mouse             +popupwin          +tag_binary        +viminfo
+autocmd           +cursorbind        -hangul_input      -mouseshape        +postscript        -tag_old_static    +vreplace
+autochdir         +cursorshape       +iconv             +mouse_dec         +printer           -tag_any_white     +wildignore
-autoservername    +dialog_con        +insert_expand     -mouse_gpm         +profile           -tcl               +wildmenu
-balloon_eval      +diff              +ipv6              -mouse_jsbterm     -python            +termguicolors     +windows
+balloon_eval_term +digraphs          +job               +mouse_netterm     -python3           +terminal          +writebackup
-browse            -dnd               +jumplist          +mouse_sgr         +quickfix          +terminfo          -X11
++builtin_terms    -ebcdic            +keymap            -mouse_sysmouse    +reltime           +termresponse      -xfontset
+byte_offset       +emacs_tags        +lambda            +mouse_urxvt       +rightleft         +textobjects       -xim
+channel           +eval              +langmap           +mouse_xterm       -ruby              +textprop          -xpm
+cindent           +ex_extra          +libcall           +multi_byte        +scrollbind        +timers            -xsmp
-clientserver      +extra_search      +linebreak         +multi_lang        +signs             +title             -xterm_clipboard
[...]

PR #1269, opened 2024-03-03
Fix PlugClean error when the default branch has changed

### LINKED ISSUE #1253: PlugClean fails to recover when remote renamed default branch
When [vim-gitgutter](https://github.com/airblade/vim-gitgutter) changed the remote, default branch from `master` to `main` a while ago, my calls to `:PlugClean` started to fail with 

```
Error detected while processing command line..function <SNR>2_clean[13]..<SNR>2_git_validate:
line   36:
E688: More targets than List items
```

The error occurs [here](https://github.com/junegunn/vim-plug/blob/ca0ae0a8b1bd6380caba2d8be43a2a19baf7dbe2/plug.vim#L2349).

Running `git status` in the vim-gitgutter's worktree shows
```
On branch master
Your branch is based on 'origin/master', but the upstream is gone.
  (use "git branch --unset-upstream" to fixup)

nothing to commit, working tree clean
```

The result from the `system` call on the vim-plug quoted line above is:
```
['''git <command> [<revision>...] -- [<file>...]''']
```

The underlying issue triggering this behavior is the outdated symbolic ref `refs/remotes/origin/HEAD`. This does not get updated on a git fetch. A `git remote set-head origin -a` can do the trick. But this does involve another network call to query the remote.

I run `PlugClean!` daily as part of [this setup script](https://github.com/kaihowl/dotfiles/blob/a01f529f8b6a938cf33369df457bb0faec999e65/nvim/install.sh#L40).

https://github.com/kaihowl/dotfiles/blob/a01f529f8b6a938cf33369df457bb0faec999e65/nvim/config--nvim--init.vim.symlink#L100-L167

```
NVIM v0.9.2
Build type: Release
LuaJIT 2.1.1692716794

   system vimrc file: "$VIM/sysinit.vim"
  fall-back for $VIM: "/usr/local/share/nvim"

Run :checkhealth for more info
```

- Type:
    - [?] Bug
    - [?] Enhancement
- OS:
    - [ ] All/Other
    - [ ] Linux
    - [x] macOS
    - [ ] Windows
- Vim:
    - [ ] Terminal Vim
    - [ ] GVim
    - [x] Neovim

junegunn:
Fix #1253

### DIFF
--- a/plug.vim
+++ b/plug.vim
@@ -2360,18 +2360,21 @@ function! s:git_validate(spec, check_branch)
               \ current_branch, origin_branch)
       endif
       if empty(err)
-        let [ahead, behind] = split(s:lastline(s:system([
-        \ 'git', 'rev-list', '--count', '--left-right',
-        \ printf('HEAD...origin/%s', origin_branch)
-        \ ], a:spec.dir)), '\t')
-        if !v:shell_error && ahead
-          if behind
+        let ahead_behind = split(s:lastline(s:system([
+          \ 'git', 'rev-list', '--count', '--left-right',
+          \ printf('HEAD...origin/%s', origin_branch)
+          \ ], a:spec.dir)), '\t')
+        if v:shell_error || len(ahead_behind) != 2
+          let err = "Failed to compare with the origin. The default branch might have changed.\nPlugClean required."
+        else
+          let [ahead, behind] = ahead_behind
+          if ahead && behind
             " Only mention PlugClean if diverged, otherwise it's likely to be
             " pushable (and probably not that messed up).
             let err = printf(
                   \ "Diverged from origin/%s (%d commit(s) ahead and %d commit(s) behind!\n"
                   \ .'Backup local changes and run PlugClean and PlugUpdate to reinstall it.', origin_branch, ahead, behind)
-          else
+          elseif ahead
             let err = printf("Ahead of origin/%s by %d commit(s).\n"
                   \ .'Cannot update until local changes are pushed.',
                   \ origin_branch, ahead)


junegunn:
I was able to trigger the code by removing `.git/refs/heads/master`.

```
x vim-xxx:
    Failed to compare with the origin. The default branch might have changed.
    PlugClean required.
```

junegunn:
See #1005 for the discussion on automatic update of the default branch.

adamency:
@junegunn Hi. I have read the threads, #1005 and #1272, but this fails to answer my question.

Can you please clarify this error message ?

```
"Failed to compare with the origin. The default branch might have changed.\nPlugClean required."
```

As per your own documentation:

```
`PlugClean[!]`                       | Remove unlisted plugins (bang version will clean without prompt)
```

but your error message tells about using `PlugClean` in the context of **installed** plugins, so either your code has changed and `PlugClean` is doing something else/more now, or your error message is wrong or missing some explanation.

Basically, my comment boils down to the fact that I was running `PlugStatus` to check the status of my plugins after the update of vim-plug itself via `PlugUpgrade` (which was supposed to solve the default branch issue), when I got this error message for all my installed plugins. However I do **not** want to remove the unlisted plugins from the filesystem and I still want to correct the `PlugStatus`.

Please provide information to better understand the issue and solve it.

junegunn:
@adamency 

> when I got this error message for all my installed plugins.

All plugins? That doesn't sound right. Please open a new issue explaining the problem, with the plug section of your Vim configuration, screenshots, etc.

Issue #1297, opened 2024-10-11
:PlugStatus reports 'Diverged from origin/master ..' when plugin installed from tag


hrls:
After fresh `:PlugInstall` with plugin from git tag, next `:PlugStatus` reports this error.
`:PlugClean` do not help as well as manual clean and installing again.
<img width="1256" alt="image" src="https://github.com/user-attachments/assets/ccbe71fa-44e6-4a89-bb7b-345e08187546">

------------------------------

```
Plug 'nvim-lua/plenary.nvim'
Plug 'nanotech/jellybeans.vim' " colorscheme
Plug 'nvim-telescope/telescope.nvim', { 'tag': '0.1.8' }
Plug 'dense-analysis/ale'

```

```
:version                                                                                                                                                         
NVIM v0.10.2                                                                                                                                                      
Build type: Release                                                                                                                                               
LuaJIT 2.1.1727870382                                                                                                                                             
Run ":verbose version" for more info 
```

- Type:
    - [x] Bug
- OS:
    - [x] macOS
- Vim:
    - [x] Neovim

EESchneider:
A workaround for this is to instead specify the commit the tag is associated with. For telescope this means `Plug 'nvim-telescope/telescope.nvim', { 'commit': 'a0bbec2' }`

PR #1321, opened 2025-11-04
Fix PlugStatus reporting divergence when a tag is specified

EESchneider:
Fixes https://github.com/junegunn/vim-plug/issues/1297, where plugins specified with `Plug xxx, {'tag': yyy}` are falsely detected by `:PlugStatus` as needing cleanup.


junegunn:
Merged, thanks!

PR #1326, opened 2026-02-21
Fix Support for PowerShell under Windows with VIM 8+

tiamat18:
The following 'vim-plug' commands would fail and/or display errors when executed on by VIM 8+ on a Windows platform with the 'shell' option set to 'powershell' or 'pwsh':
  - PlugInstall
  - PlugUpdate
  - PlugClean
  - PlugStatus
  - PlugDiff

There were two causes for these errors:
  - A bug in VIM itself (resolved by patch 9.2.6) with how compound PowerShell commands are handled by the 'system()' vimscript function.
  - A bug in the 's:vim8' branch of the private 'vim-plug' function 's:spawn()' where the actual command to be executed was escaped for 'cmd.exe' but the 'cd' prefix added by 's:with_cd()" was escaped for PowerShell.

Describe the details of your PR ...

### DIFF
--- a/plug.vim
+++ b/plug.vim
@@ -997,7 +997,7 @@ function! s:bang(cmd, ...)
     let [sh, shellcmdflag, shrd] = s:chsh(a:0)
     " FIXME: Escaping is incomplete. We could use shellescape with eval,
     "        but it won't work on Windows.
-    let cmd = a:0 ? s:with_cd(a:cmd, a:1) : a:cmd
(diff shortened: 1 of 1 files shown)


junegunn:
I'm not a Windows user so I cannot reproduce the problem and verify the fix, but the patch itself looks solid. Let me know if you think it's ready for merge.

junegunn:
Merged, thanks!

PR #1339, opened 2026-08-16
Add 'onclean' pre-clean hook option

chudur-budur:
Lets a plugin run a shell command, Vim command, or funcref right before `:PlugClean` scans and removes plugin directories, mirroring the existing `'do'` post-update hook.

Now you can do things like this:

```vim
Plug 'chudur-budur/mlir-vim', { 'do': './install.py', 'onclean': './uninstall.sh' }
```

It persists the state of plugins in the `autoload/.onclean_state` file so that whenever a plugin installation line with this hook is commented out, it will still know which one it's going to call the hook on. i.e. -- 

```vim
" Plug 'chudur-budur/mlir-vim', { 'do': './install.py', 'onclean': './uninstall.sh' }
```

will run `./uninstall.sh` upon `:PlugClean` command, even if after the line is commented out.

### DIFF
--- a/README.md
+++ b/README.md
@@ -274,6 +274,7 @@ vim.cmd('silent! colorscheme seoul256')
 | `dir`                   | Custom directory for the plugin                             |
 | `as`                    | Use different name for the plugin                           |
 | `do`                    | Post-update hook (string or funcref)                        |
+| `onclean`               | Pre-clean hook (string or funcref)                           |
 | `on`                    | On-demand loading: Commands or `<Plug>`-mappings            |
 | `for`                   | On-demand loading: File types                               |
 | `frozen`                | Do not remove and do not update unless explicitly specified |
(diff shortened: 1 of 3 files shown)


junegunn:
Hi, I've been very conservative about adding new functionality to vim-plug to keep it simple.

I assume the main motivation for working on this feature is to handle the install and uninstall scripts of your Vim plugin.

- https://github.com/chudur-budur/mlir-vim/blob/main/install.sh
- https://github.com/chudur-budur/mlir-vim/blob/main/uninstall.sh

But looking at those scripts, I can tell they are unnecessary. vim-plug, or any other plugin manager, makes sure that those directories (ftdetect, indent, syntax, etc) are recognized by Vim (via `&runtimepath` manipulation), so there's absolutely no reason to copy those files manually. I think you should not instruct your users to run that install script for better experience.

